Issue #12, opened 2019-12-31
Download page is confusing


hasufell:
https://www.haskell.org/downloads/ is confusing

* it doesn't concisely present the most popular options, but instead directs to haskell-platform, which then might direct to your packager manager, ghcup or other things, which creates another layer of confusing indirection
* "minimal installer" is a duplicate of haskell-platforms redirection to system package managers
* the description of haskell-platform "Installs GHC, Cabal, and some other tools, along with a starter set of libraries in a global location on your system." is inaccurate for choices like ghcup, which does not install in a global location
* it doesn't direct to chocolatey, which is considered the most properly packaged GHC for windows, according to some, although its popularity is surprisingly low according to the haskell 2019 survey (which might be a good reason to have it on the download page)

mattbaumann:
I agree, it is very confusing, probably it ballooned over time a bit.

Suggested Layout:

1. Contents
2. Platform Installer Links (for beginners) + chocolatey
3. Advanced Installation
  1. Stack
  2. Additional Libraries
4. Need Help? -> Link to Help Page.

Following sections should be deleted:
1. Minimal Installer
2. Where you can get help (should get own page)

Unfortunately is the https://wiki.haskell.org/Distributions Page, which is the first section  another of those broken pages. It is a page about _IDE's_ unlike the title suggests wrongfully. It does not mention IntelliJ for example. This page has to be refactored too.

hasufell:
> Platform Installer Links (for beginners) + chocolatey

Do we still need the Haskell Platform links if windows users should prefer to use Chocolatey? For linux and BSD, there are distro packages, binary distribution (the tarballs) and ghcup.

> Advanced Installation

What would this section be?

> Stack

This is an advanced tool like nix that can be used with haskell. Albeit its popularity, does it belong on the download page or should this rather be linked to as a project management tool?

mattbaumann:
I agree with you on this one, hmm. Unfortunately, as you can see there is not much traffic on this repository, so there is not really an open group discussion here, except us.

Why not, I'll produce a pull request in the upcoming days with your and mine changes, if that's alright, and normally people start a discussion as soon as there is an request. So hang on and keep the discussion up as soon as my pull request is available. Thanks for the suggestion by the way.

AMDphreak:
> I agree with you on this one, hmm. Unfortunately, as you can see there is not much traffic on this repository, so there is not really an open group discussion here, except us.
> 
[...]

PR #14, opened 2020-01-23
Download Page: Rearrangement of sections

mattbaumann:
Solves #12. This pull-request is a draft, as I would like to consider your comments as early as possible. Upon your comments, I'll make further changes until the final version is available.

* Sections are rearranged in order to be more "beginner" friendly by providing an order of importance and separating the rest of the page with "Advanced" to show what option would be the preferred one. 

* Removing double information (Platform and Minimal Installer were the same tool).

### DIFF
--- a/downloads.markdown
+++ b/downloads.markdown
@@ -6,48 +6,40 @@ isDownloads: true
 
 # Downloads
 
-There are three widely used ways to install the Haskell toolchain on supported platforms. These are:
+## Haskell Platform
 
-*   [Minimal installers](#minimal): Just GHC (the compiler), and build tools (primarily Cabal and Stack) are installed globally on your system, using your system's package manager.
+### What it is
 
-*   [Stack](#stack): Installs the `stack` command globally: a project-centric build tool to automatically download and manage Haskell dependencies on a project-by-project basis.
+<a name="platform"></a>The Haskell Platform is a self-contained, all-in-one installer. After download, you will have everything necessary to build Haskell programs against a core set of useful libraries. It comes in both core versions with tools but no libraries outside of GHC core, or full versions, which include a broader set of globally installed libraries.
 
-*   [Haskell Platform](#platform): Installs GHC, Cabal, and some other tools, along with a starter set of libraries in a global location on your system.
+### What you get
 
-These options make different choices as to what is installed globally on your system and what is maintained in project-specific environments. Global installations allow more sharing across users and projects, but at the cost of potential conflicts between projects. To avoid these conflicts, each option has a lightweight _sandboxing_ feature that creates largely self-contained, per-project environments. With Minimal you can optionally sandbox the libraries, avoiding most conflicts. Stack sandboxes the compiler, tools and libraries, so avoids nearly all kinds of conflicts between projects. With Platform you can also optionally sandbox libraries, but not the globally installed platform libraries.
+*   The [Glasgow Haskell Compiler](https://www.haskell.org/ghc)
+*   The [Cabal build system](https://www.haskell.org/cabal/), which can install new packages, and by default fetches from [Hackage](https://hackage.haskell.org/), the central Haskell package repository.
+*   the [Stack](http://docs.haskellstack.org) tool for developing projects
+*   Support for profiling and code coverage analysis
+*   35 core & widely-used [packages](https://www.haskell.org/platform/contents.html)
+
+### How to get it
+
+The Platform is provided as a single installer, and can be downloaded at the links below.
+
+*   [Linux](http://www.haskell.org/platform/linux.html)
+*   [OS X](http://www.haskell.org/platform/mac.html)
+*   [Windows](http://www.haskell.org/platform/windows.html)
+
+* * *
 
 ### Haskell IDEs & other distributions
 
 In addition to the generic, cross-platform Haskell toolchain described above, there are also easy-to-use, platform-specific distributions and IDEs. The Haskell Wiki contains a [list of the most popular ones](https://wiki.haskell.org/Distributions).
 
+
 * * *
 
-## Minimal installers
+## Advanced Installation with Stack
 
-### What they are<a name="minimal"></a>
-
-Minimal installers provide centrally the [GHC](https://www.haskell.org/ghc) compiler and the [Cabal](https://www.haskell.org/cabal/) and [Stack](https://github.com/commercialhaskell/stack) tools for installing packages. Some may install further build tools (i.e. for parsing and lexing) as well.
-
-### What you get
-
-*   Only the core libraries necessary for each platform are included.
-*   Cabal or Stack must be used to download and install packages after installation.
-
-### How to get them
-
-*   [Linux](/downloads/linux)
-*   [OS X](https://www.haskell.org/platform/mac.html) (via the core platform)
-*   [Windows](https://www.haskell.org/platform/windows.html) (via the core platform)
-
-### Where to get help<a name="help"></a>
-
-*   For help learning Haskell itself, start with the [Documentation](https://www.haskell.org/documentation) page on the [Haskell Wiki](https://wiki.haskell.org/).
-*   If you need help with [GHC](https://www.haskell.org/ghc)---the Haskell compiler---there is a comprehensive [GHC User Manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/index.html).
-*   For help using Cabal to download or create additional packages (see [below](#libraries)), there is the [Cabal User Guide](https://www.haskell.org/cabal/users-guide/).
-*   For help using Stack to download or create packages, see the stack documentation [below](#stackhelp).
-*   Finally, you can ask questions of other Haskell users and experts on the [#haskell IRC channel](irc://irc.freenode.net/haskell) on the Freenode IRC network.
-
-## Stack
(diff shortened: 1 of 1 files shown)


downloads.markdown
```diff
@@ -6,48 +6,40 @@ isDownloads: true
...
 # Downloads
 
-There are three widely used ways to install the Haskell toolchain on supported platforms. These are:
-
-*   [Minimal installers](#minimal): Just GHC (the compiler), and build tools (primarily Cabal and Stack) are installed globally on your system, using your system's package manager.
-
-*   [Stack](#stack): Installs the `stack` command globally: a project-centric build tool to automatically download and manage Haskell dependencies on a project-by-project basis.
-
-*   [Haskell Platform](#platform): Installs GHC, Cabal, and some other tools, along with a starter set of libraries in a global location on your system.
+## Haskell Platform
```
hasufell, line 9: There exists no haskell platform anymore.

For linux and mac, it redirects to either 1. package managers or 2. ghcup. This is another layer of indirection causing confusion for users.

For windows it redirects to chocolatey. And windows is an even more complicated matter being discussed: https://gitlab.haskell.org/ghc/ghc/issues/18104

All this information should be **right here**, not on the haskell platform page.

downloads.markdown
```diff
@@ -6,48 +6,40 @@ isDownloads: true
...
 
-### What you get
+* * *
 
-*   Only the core libraries necessary for each platform are included.
-*   Cabal or Stack must be used to download and install packages after installation.
+### Haskell IDEs & other distributions
 
-### How to get them
+In addition to the generic, cross-platform Haskell toolchain described above, there are also easy-to-use, platform-specific distributions and IDEs. The Haskell Wiki contains a [list of the most popular ones](https://wiki.haskell.org/Distributions).
```
hasufell, line 35: The wiki page is outdated, is it not? I don't see any information about modern IDEs. "Haskell for mac" runs ghc-8.6.4, etc.

downloads.markdown
```diff
@@ -6,48 +6,40 @@ isDownloads: true
...
 
-### Where to get help<a name="help"></a>
+* * *
 
-*   For help learning Haskell itself, start with the [Documentation](https://www.haskell.org/documentation) page on the [Haskell Wiki](https://wiki.haskell.org/).
-*   If you need help with [GHC](https://www.haskell.org/ghc)---the Haskell compiler---there is a comprehensive [GHC User Manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/index.html).
-*   For help using Cabal to download or create additional packages (see [below](#libraries)), there is the [Cabal User Guide](https://www.haskell.org/cabal/users-guide/).
-*   For help using Stack to download or create packages, see the stack documentation [below](#stackhelp).
-*   Finally, you can ask questions of other Haskell users and experts on the [#haskell IRC channel](irc://irc.freenode.net/haskell) on the Freenode IRC network.
+## Advanced Installation with Stack
```
hasufell, line 40: IMO, stack doesn't belong here. It belong to the wiki page: https://wiki.haskell.org/Distributions

downloads.markdown
```diff
@@ -159,3 +122,12 @@ $$ git clone https://github.com/haskell/network
...
 ```
+# You need Help?
+
+### Where to get help<a name="help"></a>
+
+*   For help learning Haskell itself, start with the [Documentation](https://www.haskell.org/documentation) page on the [Haskell Wiki](https://wiki.haskell.org/).
+*   If you need help with [GHC](https://www.haskell.org/ghc)---the Haskell compiler---there is a comprehensive [GHC User Manual](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/index.html).
+*   For help using Cabal to download or create additional packages (see [below](#libraries)), there is the [Cabal User Guide](https://www.haskell.org/cabal/users-guide/).
+*   For help using Stack to download or create packages, see the stack documentation [below](#stackhelp).
+*   Finally, you can ask questions of other Haskell users and experts on the [#haskell IRC channel](irc://irc.freenode.net/haskell) on the Freenode IRC network.
```
hasufell, line 133: IMO, it would be better to have the IRC link be a link to http://webchat.freenode.net/?randomnick=1&channels=%23haskell&uio=d4

downloads.markdown
gbaz: I disagree. We should point to the raw binaries, and then just point to the platform page for the recommended installation methods for each system.

downloads.markdown
gbaz: if the wiki page is outdated, people should edit it. its a wiki!

gbaz:
Note: people that want to view the rendered markdown directly for ease of reading can just follow this link: https://github.com/haskell-infra/www.haskell.org/blob/2ae635a9bc1b7229f1b3e5ac7c44e3018339a9a1/downloads.markdown

downloads.markdown
gbaz: The section can be simplified by saying that the platform points to recommended installers for common systems, and leave it at that.

downloads.markdown
TikhonJelvis: How can we ensure that people are consistently keeping the Wiki up to date? This seems difficult, especially since getting a Wiki account is non-trivial.

gbaz: getting a wiki account is trivial. you just email and ask for it. i see a few accounts added each day.

there is no way to ensure the wiki is kept up to date. but there is no way to ensure anything else is kept up to date either! people either do it, or they don't, and then content becomes stale. this is true no matter what.

gbaz: (one caveat: if you tell people nothing, then this never becomes out of date, but then... you have told them nothing.)

downloads.markdown
hasufell: @gbaz I do not understand your remarks. There are no binaries on the haskell platform page. The term "haskell platform" by now is quite confusing, because it doesn't exist. It just redirects the user further.

downloads.markdown
hasufell: > if the wiki page is outdated, people should edit it. its a wiki!

It is not a public wiki. I'm trying to find the discussion, but very recently a user reported to have requested an account 3+ months ago and never got any reply.

gbaz: The wiki is public, but simply requires a registration step. The barrier to entry is "send an email". The only alternative is for it to be filled by spam bots. If a user requested an account and didn't get one, their email was lost. I see requests for accounts filled efficiently almost every day.

downloads.markdown
gbaz: What I'm suggesting is that we can point to ghc's raw binary downloads somewhere on this page, so people are aware of them. Independently, we can also point to the platform and all we need say about it is it collects recommended installers for common systems.

I am suggesting that extra hop is _fine_, since the platform "brand" can then stay uniform across different systems. It is just that the extra hop can now be significantly simplifed, to make this section very short and clear.

"Raw binaries of ghc are available from [..]. Most people will want to use an installation system to manage ghc and related tools such as cabal. The haskell platform collects recommended installation methods across common platforms [...]." Done.

hasufell: > What I'm suggesting is that we can point to ghc's raw binary downloads somewhere on this page, so people are aware of them. Independently, we can also point to the platform and all we need say about it is it collects recommended installers for common systems.

If you are referring to the bindists, then I'm pretty sure most novice users will not be able to install these. That is basically one of the problems ghcup tried to resolve and it's still getting lots of bug reports from mac users, who don't know how to adjust their PATH variable. Download, unpacking, dealing with configure flags, custom locations, build requirements, ...is **advanced** use. If it should be mentioned on the landing page, then in a way that makes it clear for users this isn't something they want generally, unless they know what they are doing.

Otherwise we'll have to deal with more confused users in `#haskell`, I am pretty sure.

> I am suggesting that extra hop is fine, since the platform "brand" can then stay uniform across different systems. It is just that the extra hop can now be significantly simplifed, to make this section very short and clear.

And I am strongly against this. The entire reason of the issue I raised at #12 was that this extra hop is confusing and the branding "haskell platform" confuses users further.

You're landing on the download page and the download page tells you "To get to the download page, click here, but we call it 'haskell platform', because of legacy reasons, when it was still an installer."

downloads.markdown
hasufell: > The only alternative is for it to be filled by spam bots.

There are technical solution to this, such as using a wiki that supports github OAuth. haskell gitlab also uses that model.

downloads.markdown
mattbaumann: What should I do? Remain Installer in place or remove it?

My suggestion is even more bold. Integrate Haskell-Platform site to downloads and create 'Advanced' section with the other options. How's that?

mattbaumann:
At a second glance, I'm actually for a bold action with following page content:

1. Haskell Platform Page integrated as a section
2. Advanced Downloads see Wiki
3. Help?

Not only would that resolve all issues, it allows greater participation, since Wiki content is easier to replace than the website stuff. Also, the 'haskellers' are actually used to browse the wiki and don't mind the redirection.

downloads.markdown
hasufell: More proof that haskell platform both as branding and subpage is confusing users: https://www.reddit.com/r/haskell/comments/guecax/confused_by_the_haskell_installation_options

@mattbaumann my opinion is to remove all mentions of "haskell platform" as well as stack and provide a concise list of options per platform. Whether that follows the design of current haskell platform or not doesn't matter much and should be up to whomever provides that PR.

hasufell:
alternative PR https://github.com/haskell-infra/www.haskell.org/pull/71

downloads.markdown
ketzacoatl: FTR, this wiki is not as open as you suggest gbaz.

image

This prevents correcting/updating the page to keep it in sync with reality.

We should do away with the wiki and let people host their own content and/or combine good, maintained content into the haskell.org website.

gbaz: That is a single protected page (there are a few) on a generally open to edit wiki. The reason that page is protected is by request -- people were concerned that since it contained authoritative information (including a link to donations) that it be specially protected from vandalism. Conversations on an old PR are not a place to have further discussion about plans for the wiki.

TikhonJelvis:
Let's close this in favor of #71. If there are any changes we want to keep from here, it would be easier to review with a smaller, more focused PR.

Thanks!

PR #19, opened 2020-03-29
chocolatey: Add instructions for chocolatey

Mistuke:
This replaces the Installation instruction for the `Minimal` installation for Windows with `Chocolatey` instructions instead of `Haskell-Platform`.

It also adds some basic usage instructions and how to move from `Haskell-Platform` to `Chocolatey`.

### DIFF
--- a/downloads/windows.markdown
+++ b/downloads/windows.markdown
@@ -1,0 +1,106 @@
+---
+title: Downloads for Windows
+page: downloads
+isDownloads: true
+---
+
+# Downloads for Windows
+
+## Package-based install
+
+The package based installers on Windows use [chocolatey](https://chocolatey.org).
+To use them first [configure Chocolatey](https://chocolatey.org/install) on your machine.
+
+### Setup new environment
+
+Steps to setup ghc and cabal are given in the [ghc chocolatey](https://chocolatey.org/packages/ghc)
+
+For new installations it is recommended to use `Haskell-Dev` which will also setup and configure `MSys2` for use with `GHC` and `Cabal`.
+
+**If upgrading from `Haskell-Platform` please see instructions below.**
+
+`Haskell-Dev` can be installed from Chocolatey as follows from an elevated command prompt:
+
+```bash
+choco install haskell-dev
+refreshenv
+```
+
+Chocolatey has limited [non-administrative](https://chocolatey.org/docs/installation#non-administrative-install) install support for those that absolutely need it.
+
+After this you can upgrade or install additional `ghc` independently (see below).
+
+Read more about [haskell-dev](https://hub.zhox.com/posts/introducing-haskell-dev/) or about how to integrate with Cloud CI systems.
+
+Packages are installed into `%ProgramData%\chocolatey\lib\ghc\` and `%ProgramData%\chocolatey\lib\cabal\`
+
+Steps to setup stack are [on the stack website](https://docs.haskellstack.org/en/stable/install_and_upgrade/#ubuntu).
+
+### Upgrading from Haskell-Platforms
+
+If moving from a `Haskell-Platform` based installation to `Chocolatey` you will need to clean your `cabal` configuration before uninstalling `Haskell-Platform`.
+
+To do this run:
+```bash
+cabal user-config init -f
+```
+
+and proceed to uninstall `Haskell-Platform` before installing using Chocolatey otherwise the Chocolatey packages may not modify the Cabal configuration file.
+
+Do not keep both `Haskell-Platform` and `Chocolatey` based set up at the same time as they modify the same global configuration.
+
+### Upgrading/Downgrading GHC or Cabal
+
+`GHC` or `Cabal` can be upgraded using the [Chocolatey upgrade](https://chocolatey.org/docs/commandsupgrade) command.
+
+For `GHC` (which may or may not trigger a cabal upgraded based on the version constraints)
+
+```bash
+choco upgrade ghc
+refreshenv
+```
+
+For just `cabal` which will never trigger a `ghc` upgrade.
+```bash
+choco upgrade cabal
+refreshenv
+```
+
+For both `ghc` and `cabal`
+
+```bash
+choco upgrade ghc cabal
+refreshenv
+```
+
+Downgrades can be performed by specifying the version number during installation. e.g:
+
+```bash
+choco install ghc --version 8.6.5
+refreshenv
+```
+
+A list of versions available can be seen on the [ghc Chocolatey](https://chocolatey.org/packages/ghc) website or by
+using the command
+
+```bash
+choco list ghc --all
+```
+
+### Installing multiple versions of GHC
+
+Multiple versions of `GHC ` can be installed concurrently by using the `-m` switch:
+
+```bash
+choco install ghc --version 8.4.1
+choco install ghc --version 8.2.2 -m
+refreshenv
+```
+
+You can then select which one you want to use by using the full compiler name, e.g `ghc-8.4.1.exe`.
+These can be passed along to `cabal` as well:
+
+```bash
+cabal build -w ghc-8.4.1.exe
+cabal build -w ghc-8.2.2.exe
+```
--- a/downloads/windows.markdown
+++ b/downloads/windows.markdown
@@ -11,11 +11,11 @@ isDownloads: true
 The package based installers on Windows use [chocolatey](https://chocolatey.org).
 To use them first [configure Chocolatey](https://chocolatey.org/install) on your machine.
 
-### Setup new environment
+### Setting up a new environment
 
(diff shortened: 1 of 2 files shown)


gbaz:
I've updated the haskell platform site itself to now point to chocolatey for windows. This PR is still a good idea imho.

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+---
+
+# Downloads for Windows
+
+## Package-based install
+
+The package based installers on Windows use [chocolatey](https://chocolatey.org).
+To use them first [configure Chocolatey](https://chocolatey.org/install) on your machine.
+
+### Setup new environment
```
emilypi, line 14:
```suggestion
### Setting up new environment
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+Steps to setup ghc and cabal are given in the [ghc chocolatey](https://chocolatey.org/packages/ghc)
```
emilypi, line 16:
```suggestion
Follow the steps for setting up ghc and cabal given by [ghc chocolatey](https://chocolatey.org/packages/ghc)
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+For new installations it is recommended to use `Haskell-Dev` which will also setup and configure `MSys2` for use with `GHC` and `Cabal`.
```
emilypi, line 18:
```suggestion
For new installations, it is recommended to use `Haskell-Dev` which will also setup and configure `MSys2` for use with `GHC` and `Cabal`.
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+`Haskell-Dev` can be installed from Chocolatey as follows from an elevated command prompt:
+
+```bash
+choco install haskell-dev
+refreshenv
+```
+
+Chocolatey has limited [non-administrative](https://chocolatey.org/docs/installation#non-administrative-install) install support for those that absolutely need it.
+
+After this you can upgrade or install additional `ghc` independently (see below).
```
emilypi, line 31:
```suggestion
After this you can upgrade or install additional `ghc` versions independently (see below).
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+Read more about [haskell-dev](https://hub.zhox.com/posts/introducing-haskell-dev/) or about how to integrate with Cloud CI systems.
```
emilypi, line 33:
```suggestion
For integrating with Cloud CI systems, see [haskell-dev](https://hub.zhox.com/posts/introducing-haskell-dev/). 
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+Packages are installed into `%ProgramData%\chocolatey\lib\ghc\` and `%ProgramData%\chocolatey\lib\cabal\`
```
emilypi, line 35: This is a Chocolatey fact. We probably don't need to repeat this here.

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+### Upgrading from Haskell-Platforms
+
+If moving from a `Haskell-Platform` based installation to `Chocolatey` you will need to clean your `cabal` configuration before uninstalling `Haskell-Platform`.
+
+To do this run:
+```bash
+cabal user-config init -f
+```
+
+and proceed to uninstall `Haskell-Platform` before installing using Chocolatey otherwise the Chocolatey packages may not modify the Cabal configuration file.
```
emilypi, line 48:
```suggestion
and proceed to uninstall `Haskell-Platform` before installing using Chocolatey. Otherwise, the Chocolatey packages may not modify the Cabal configuration file.
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+
+Do not keep both `Haskell-Platform` and `Chocolatey` based set up at the same time as they modify the same global configuration.
```
emilypi, line 50:
```suggestion
Do not attempt to use both `Haskell-Platform` and `Chocolatey`-based installations at the same time, as they modify the same global configuration.
```

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Downgrades can be performed by specifying the version number during installation. e.g:
+
+```bash
+choco install ghc --version 8.6.5
+refreshenv
+```
+
+A list of versions available can be seen on the [ghc Chocolatey](https://chocolatey.org/packages/ghc) website or by
```
emilypi, line 83:
```suggestion
A list of available versions can be found on the [GHC Chocolatey](https://chocolatey.org/packages/ghc) website or by
```

downloads/windows.markdown
Mistuke: I've done so because the `Linux` instructions seem to explicitly point out where they are installed. Note that while this is the chocolatey default it's up to packages to decide where they're installed.

Mistuke:
Thanks @emilypi I can remove the line about the install directory but just wanted to be consistent with the instructions for Hvr's PPA.

@gbaz thanks! I noticed that there are a lot of white spaces before the `-f` in step 2 that makes it a bit easy to miss at https://www.haskell.org/platform/windows.html not sure where those are coming from.

gbaz:
Fixed, thanks. auto-wrapping + pre tags :-)

Mistuke:
Hi @gbaz , @emilypi ,

So following the recent threads on /r/haskell and haskell-cafe I am wondering if using the commandline is too hard for a certain segment of the Windows user base.

I'm not sure what to do about this one since there seems to be a lot of people stuck in the past but perhaps for them an exe based setup should remain.  Thoughts?

gbaz:
I think in one of those threads you  suggested we could wrap up  the commandline instructions in a little executable.  That sounds like a good idea to me, if someone is willing to write it! Certainly overall the NSIS installers are not nearly as good as chocolatey, and we should  go forwards, not backwards.

emilypi:
I agree with that approach @gbaz. Unfortunately I can't be the person to help with that, because I don't have a windows machine, but yeah definitely 👍

Mistuke:
Alright, had a chat with @bgamari and we came up with this approach https://gitlab.haskell.org/ghc/ghc/issues/18104 which we're looking into.

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+# Downloads for Windows
+
+## Package-based install
+
+The package based installers on Windows use [chocolatey](https://chocolatey.org).
+To use them first [configure Chocolatey](https://chocolatey.org/install) on your machine.
+
+### Setting up a new environment
+
+Follow the steps for setting up ghc and cabal given by [ghc chocolatey](https://chocolatey.org/packages/ghc)
```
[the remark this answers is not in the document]
Mistuke: I think that's fair, but you only need `haskell-dev` for the initial install.

downloads/windows.markdown
```diff
@@ -0,0 +1,106 @@
...
+A list of available versions can be found on the [GHC Chocolatey](https://chocolatey.org/packages/ghc) website or by
+using the command
+
+```bash
+choco list ghc --all
+```
+
+### Installing multiple versions of GHC
+
+Multiple versions of `GHC ` can be installed concurrently by using the `-m` switch:
```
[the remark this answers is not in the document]
Mistuke: `-m` is short for `Multiple` it just lets the package manager allow more than one version of the same package to be installed at the same time.  Normally this would result in an error saying you have a more recent/older version install and to upgrade or downgrade instead..

Mistuke:
Sorry for the late response @TikhonJelvis , been working hard on finishing things for GHC 8.12.

I guess this PR should be parked for a bit since after the whole dust up by people I have spoken with @bgamari and we have settled on writing an msi wrapper around chocolatey.  These instructions would essentially only be for power users then.

TikhonJelvis:
Sounds good. An installer sounds like the most intuitive option for Windows users, and I'm glad you're working on it!

ketzacoatl:
@TikhonJelvis / @emilypi is there a next-step with this PR? Should it be merged, closed, or something else?

downloads/windows.markdown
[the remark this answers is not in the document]
jneira: I would use one of the long aliases for that Param: `--allow-multiple` for example

jneira:
This contains a invaluable work and should be included for sure, in the actual URL or in another one.
Iirc the last proposal with the one click installer was a wrapper over the choco install so to have some info about it is a must.

ketzacoatl:
@rebeccaskinner / @TikhonJelvis / @jneira what do we do from here?

TikhonJelvis:
@Mistuke Going to close this PR for now. We can reopen it when we either want to get the instructions above merged or when you and the HF working group have a new Windows installation process ready.

PR #25, opened 2020-05-08
Community: added Haskell Love Conference

idabzo:


TikhonJelvis:
Thanks for the suggestion! Adding that link looks good to me. Let me find somebody with merge permissions to merge it in...

Issue #29, opened 2020-05-30
Site not building locally on Linux


TikhonJelvis:
I'm using Nix to build the site on a non-NixOS Linux system. The Nix shell runs correctly, but `buildAndWatch` fails with the following error:

```
Compiling
  [ERROR] ./community.markdown: hGetContents: invalid argument (invalid byte sequence)
```

I suspect this is related to an issue with GHC on Nix and the locale encoding (https://github.com/NixOS/nixpkgs/issues/64603).

TikhonJelvis:
I'm trying to fix this right now, but no guarantees I'll get anywhere...

TikhonJelvis:
Should work by default now that #30 has been merged.

If anyone runs into this error again, please comment here or open a new issue. Thanks!

Martinsos:
Running into this problem!

Here is the exact error I get:
```
[master=] ~/git/www.haskell.org$ ./buildAndWatch 
copying path '/nix/store/20dyxqvd9rnwzdnpdzb8vnv8w3vwa8wx-glibc-locales-2.31-74' from 'https://cache.nixos.org'...
copying path '/nix/store/lipa6p96ivgwws4hp2v4zjd7jv00p04d-bzip2-1.0.6.0.1' from 'https://cache.nixos.org'...
copying path '/nix/store/m87blzdbjqhj1dghl5l2dqqnl9dhvarz-diffutils-3.7' from 'https://cache.nixos.org'...
copying path '/nix/store/6v59a9iszrs0kxk1b79v7fasn9bwi79q-bzip2-1.0.6.0.1-bin' from 'https://cache.nixos.org'...
copying path '/nix/store/pvvw66a5xshxyc9w1hx0wb56x9avpj4j-ed-1.16' from 'https://cache.nixos.org'...
copying path '/nix/store/ax8vv1ds6l81jx8cmflx3fvcl9jdxd2w-findutils-4.7.0' from 'https://cache.nixos.org'...
copying path '/nix/store/b9j6qhmxpwfvga2bfxn1khjrv3v3qxg7-gawk-5.1.0' from 'https://cache.nixos.org'...
copying path '/nix/store/7fqqkrz6m11793cw7by8wbn2q87za8ny-gnumake-4.3' from 'https://cache.nixos.org'...
copying path '/nix/store/gdr4s4xx41n8krk9carr8amvncy8x3p3-gnused-4.8' from 'https://cache.nixos.org'...
[...]

PR #30, opened 2020-05-30
Explicitly set LOCALE_ARCHIVE and LC_ALL inside nix-shell.

TikhonJelvis:
With this change, I was able to run `buildAndWatch` from a Nix shell without running into Unicode errors (see #29 for details).

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -36,6 +36,8 @@ let
         ];
       });
     }).overrideAttrs (old: {
+      LOCALE_ARCHIVE = "${pkgs.glibcLocales}/lib/locale/locale-archive";
+      LC_ALL = "en_US.UTF-8";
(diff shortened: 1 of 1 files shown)


TikhonJelvis:
@rebeccaskinner Do you have a chance to look at this PR again? Setting `LC_ALL=C.UTF-8` is the only way I got `buildAndWatch` working consistently in a Nix shell on my (non-NixOS) Linux machine. Without this setting, Hakyll would fail build the site because of encoding issues.

TikhonJelvis:
I reproduced both the problem and the fix with `nix-shell --pure` to ensure it wasn't due to my own local settings.

emilypi (approved):

Issue #31, opened 2020-05-30
`nix-shell` warns about a non-free license


rebeccaskinner:
Error:

On a fresh clone, running `nix-shell` fails with the following error:
```
error: Package ‘haskell-org-0.1.0.0’ in /nix/store/gxsq867akcklrbvzq1ph4nsadckgipaw-cabal2nix-www.haskell.org/default.nix:3 has an unfree license (‘unfree’), refusing to evaluate.

a) For `nixos-rebuild` you can set
  { nixpkgs.config.allowUnfree = true; }
in configuration.nix to override this.

b) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
  { allowUnfree = true; }
to ~/.config/nixpkgs/config.nix.

(use '--show-trace' to show detailed location information)
```
Environment:
Nix 2.3.4 on Linux (PopOS)

Steps to reproduce:
1. Install Nix
2. Clone this repository
3. Enter a nix shell

Suggested Resolution:
Address the licensing issue so that `nix-shell` does not require enabling non-free licenses.

TikhonJelvis:
Okay, this should be resolved now.

PR #32, opened 2020-05-30
Added the license field to the .cabal file.

TikhonJelvis:
This makes the licensing of the package clear and ensures it can be used without globally configuring `allowUnfree = true` in Nix.


TikhonJelvis:
This should address #31

PR #34, opened 2020-05-30
Bumped default Nixpkgs to a rev from 20.03 and GHC to 8.8.3.

TikhonJelvis:
This ensures we're using a more recent GHC release and generally more recent packages. Apart from keeping us up to date, this should also let Nix fetch more packages from the public binary caches, saving on build time.

This required bumping the Hakyll requirement in the .cabal file to `4.13.*`.

The site still builds correctly with all of these changes.

### DIFF
--- a/haskell-org.cabal
+++ b/haskell-org.cabal
@@ -6,7 +6,7 @@ cabal-version:      >= 1.10
 executable site
   main-is:          site.hs
   build-depends:    base == 4.*
-                  , hakyll == 4.12.*
+                  , hakyll == 4.13.*
                   , time
                   , filepath
   ghc-options:      -threaded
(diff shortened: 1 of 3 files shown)


haskell-org.cabal
```diff
@@ -6,7 +6,7 @@ cabal-version:      >= 1.10
 executable site
   main-is:          site.hs
   build-depends:    base == 4.*
-                  , hakyll == 4.12.*
+                  , hakyll == 4.13.*
```
gwils, line 9: Could we loosen this to `hakyll >= 4.12 && < 4.13` ? Our code should still work on the old hakyll so I'd rather we not force a hakyll/ghc update on the build server.

TikhonJelvis: sure, let me give that a try

PR #36, opened 2020-06-06
Hot fix: haddock package name in Archlinux

huyhoang8398:


TikhonJelvis:
Thanks for the fix.

Seems like this is the same change as #33. We should get one of these merged soon :).

TikhonJelvis:
Okay, this should be fixed next time the site is deployed. Thanks for paying attention to details like this :).

PR #40, opened 2020-09-18
Remove link to spam site from IRC page

glguy:
The IRC page was quite out of date. ircbrowse.net no longer exists and is being squatted on!

### DIFF
--- a/irc.markdown
+++ b/irc.markdown
@@ -6,24 +6,15 @@ isCommunity: true
 
 # IRC Channels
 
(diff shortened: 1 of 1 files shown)


TikhonJelvis (approved):
Thanks for spotting that! Your change looks great. It's a shame to hear IRC browse is down.

Do you know if there's some alternative to IRC Browse that we can point people to? That doesn't block this PR at all, I'm just curious about that for the future.

glguy:
@TikhonJelvis tunes.org is the only consistent log source I'm aware of, unfortunately. (I know it's not searchable).

simonmichael:
It's only a few days old, but https://ircbrowse.tomsmeding.com/browse/haskell lives!

gwils (approved):

Issue #43, opened 2020-10-28
inappropriate use of url


apparluk:
![discourse](https://user-images.githubusercontent.com/52227507/97499858-e7252400-196e-11eb-8063-aacaec5b6876.png)

TikhonJelvis:
I'm sorry, what's the inappropriate use of the URL here?

apparluk:
Whom 'us' refers to in the screenshot is unspecified.

Haskell.org neither specifies Committee liability nor with the conclusion of terms where liability rests.

Consult with counsel.

PR #44, opened 2020-10-28
delete discourse content

apparluk:
For https://github.com/haskell-infra/www.haskell.org/issues/43.


gbaz:
I don't think the links to the discourse should be removed. They should stay.

PR #48, opened 2020-12-11
Make internal links jump to the header

tomjaguarpaw:
... rather than the paragraph text

The internal links to the Minimal installers and Haskell Platform sections of  https://www.haskell.org/downloads/ jump to the paragraph text below the section title, not the section title itself.  (The link to Stack doesn't but by accident rather than by design, I think. pandoc seems to add `id="stack"` to the header, which is preferred over the `<a name="stack">`).

I found this rather disorienting. When I click on an internal link I expect to go to a section header that reflects what I just clicked. I'm sure linking to the paragraph text was done to save the reader time and/or screen space but for me at least it was psychologically unsettling.

How about this PR to switch to behaviour which is, I believe, more common?

### DIFF
--- a/downloads.markdown
+++ b/downloads.markdown
@@ -22,9 +22,9 @@ In addition to the generic, cross-platform Haskell toolchain described above, th
 
 * * *
 
-## Minimal installers
+## Minimal installers<a name="minimal"></a>
 
-### What they are<a name="minimal"></a>
+### What they are
 
 Minimal installers provide centrally the [GHC](https://www.haskell.org/ghc) compiler and the [Cabal](https://www.haskell.org/cabal/) and [Stack](https://github.com/commercialhaskell/stack) tools for installing packages. Some may install further build tools (i.e. for parsing and lexing) as well.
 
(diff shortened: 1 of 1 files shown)


TikhonJelvis:
Oh, oops, didn't mean to push directly to this branch, sorry!

My suggestion was to use `{ #minimal }` instead of `<a name="minimal">`, which seems a bit neater in Pandoc.

TikhonJelvis (approved):
This change in behavior makes sense to me—jumping to the section header matches the experience I'd expect based on other sites I use.

tomjaguarpaw:
> Oh, oops, didn't mean to push directly to this branch, sorry!

That's fine. I set up the PR to allow repo owners to make changes to that branch.

TikhonJelvis:
Awesome, sounds good.

For reference, we've been operating on a two-reviewer system, so your PRs will get merged once somebody else from the Haskell.org committee reviews them. (I'll ping somebody about it.) Now that I mention it, the process we have is another thing we need to document, probably in a `CONTRIBUTING.md` file.

Also, if you're interested in helping with this sort of thing, we've got an open call for (self-)nominations to the Haskell.org committee for next year ;).

tomjaguarpaw:
> if you're interested in helping with this sort of thing, we've got an open call for (self-)nominations to the Haskell.org committee for next year ;).

Thanks for the suggestion. I have emailed the committee to self-nominate.

jaspervdj (approved):

PR #54, opened 2020-12-24
Added donation links

### LINKED ISSUE #53: Improve discoverability of findings page
It was pointed out that without a direct link, it’s hard to discover the donation page. We should consider making this a bit more visible.

decapo01:
closes #53

I made a comment in reddit about how it was difficult to find the donations area and donate.  I got a reply back to submit a PR to address this.  I had trouble getting nix and cabal to work on my mac (need to switch back to linux) so I used stack to test this out and removed all the stack files.  Let me know if there are any questions or other modifications that I need to make.

### DIFF
--- a/community.markdown
+++ b/community.markdown
@@ -8,6 +8,8 @@ isCommunity: true
 
 Haskellers interact, talk and collaborate across several mediums and around the world. There are places to learn, to teach, to ask questions, and to find contributors and collaborators.
 
+[Help support the community!](/donations/)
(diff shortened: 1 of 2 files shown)


TikhonJelvis (approved):
Thanks for the PR! Adding an extra link on the community page and an entry in the navigation menu makes sense to me.

rebeccaskinner (approved):
Looks great to me!

PR #56, opened 2021-01-06
Remove obsolete IO actions from sandbox suggestions

### LINKED ISSUE #55: IO actions not working in sandbox
When running suggested example 
```haskell
do line <- getLine; putStrLn line
```
in sandbox I get an error as below
image

This behavior seems to be common for all IO actions.

3Rafal:
This should fix #55 .
Made it similar to https://tryhaskell.org/ (which is a backend for the sandbox).


rebeccaskinner (approved):
This looks good to me

TikhonJelvis:
Thanks for this PR! I'm going to test it locally later today, then I can approve and merge.

TikhonJelvis (approved):
Looks great. Thanks for the fix!

PR #63, opened 2021-02-03
haskell-org.cabal: rename site executable --> haskell-org-site

ketzacoatl:
This ensures there are no conflicts with other hakyll sites (that might create another executable called `site`).


TikhonJelvis (approved):
Makes sense to me.

Issue #59, opened 2021-01-19
wiki vs haskell.org page content


ketzacoatl:
Hello,

While reading about the haskell.org committee, I came across:
* https://www.haskell.org/haskell-org-committee/
* and then also https://wiki.haskell.org/Haskell.org_committee

Which page is the authoritative? Should one just be a link to the other?
I find this type of duplication confusing at best and time-consuming at worst (if I want any confidence in my understanding of the material, I need to read both pages and compare the content).

I'd be happy to help with content updates, but not sure what updates would be welcome.

ketzacoatl:
Is this repo/project maintained?

TikhonJelvis:
Hey, sorry, missed your issue earlier!

I agree that the duplication is confusing. We're currently figuring out what to do with the Wiki in general, so this is something of a transitional state—although things have been moving slowly because everybody on the committee has been extra busy lately.

To me, it seems like the best solution would be to move all the relevant content about the Haskell.org Committee to the official Haskell.org website. In the past, we've gotten feedback that Wiki pages (even if they're edit protected) don't look official. Once we do that, we could replace the Wiki content itself with a link or redirect to the new page.

If you're interested in helping out, you could create a PR on this repo either adding details from the Wiki directly to the existing committee page, or creating another more detailed page with Haskell.org Committee rules/history/etc—whatever reads better. Otherwise I was thinking about creating a PR like that myself, but I won't have time to do that for a few weeks at least, so help is absolutely appreciated!

ketzacoatl:
@TikhonJelvis, that's understandable. Thank you for the guidance there, I'd be happy to at least take a look and see if I can get a PR started.

TikhonJelvis:
Awesome, thanks for looking into it!

ketzacoatl:
@TikhonJelvis it's not perfect, but it's a start. I'd like to see through updating the wiki article to point back at this page, so please review and LMK what additional updates would be best to make here. Thank you for helping to guide and move this along.

ketzacoatl:
That PR has been merged. Now we just need to update the wiki, to replace the existing content with a link to the authoritative page on haskell.org.

TikhonJelvis:
Yep. Heads up that it might take a few days—running into some technical difficulties on my end, and won't be able to deploy the site until they're fixed.

TikhonJelvis:
[...]

Issue #62, opened 2021-01-21
Committee Membership out of date?


tbm:
https://www.haskell.org/haskell-org-committee/ says

> Ryan Trinkle (term ends 2020)
> Tikhon Jelvis (term ends 2020)
> George Wilson (term ends 2020)

Should the page be updated?

TikhonJelvis:
We're a bit behind schedule on voting for new members. We'll have a vote and discussion this week and announce new members (as well as updating the page) either later this week or early next week.

TikhonJelvis:
New members were announced recently. We'll update the site first, alongside the changes in #64.

Thanks for spotting this originally, by the way :). It's easy for little bit of the site to get out of date without reminders like this.

PR #64, opened 2021-02-03
haskell-org-committee: merge in content from the wiki

ketzacoatl:
As outlined in #59, there are two pages which describe the Haskell.org committee and it's roles, responsibilities, process, history, etc. One is here on haskell.org, the other is on the wiki at https://wiki.haskell.org/Haskell.org_committee. Both pages appear authoritative, but differ in their content. At the same time, the general consensus is to move content off the wiki (particularly the more "official" content such as this). We would then replace that wiki article with a link to the this page. This is a first attempt which is more of a "bulk import" (or "merge") of the existing content. Note that there are a few places where the differences in content on the haskell.org page and wiki article may seem to be in conflict (or where one does not go as far as the other). These discrepancies and conflicts should be reviewed and updated. Some of the wiki content is more verbose than that on this page, and may still deserve some word-smithing.

A rough list of changes:
* collected the full list of links at the bottom of the page
* move the list of "committee members" a little lower in the page, bumping contact info above
* Included/merged wiki content: process, history and responsibilities content

I can make additional updates to this, but would need feedback on that. cc @tikhonjelvis

---

This is an updated PR, see https://github.com/ketzacoatl/www.haskell.org/pull/1 for the original (which was mistakenly created over there and GH doesn't let me move or update the base repo the PR targets), including @TikhonJelvis's feedback to the original.

This PR attempts to incorporate all of the feedback from that first PR.


TikhonJelvis (approved):
Looks great, thanks!

I pushed a tiny formatting tweak (changed code block to block quote), and I think it looks great to merge. I'll ping somebody else on the committee for a second review, then we can merge this, deploy the change and change the Wiki page.

TikhonJelvis:
Once this is merged, we can also do a quick PR updating the list of committee members (addressing #62). I'd like to do that in a separate PR to keep the history easy to navigate.

ketzacoatl:
That sounds great @TikhonJelvis, I look forward to seeing the updates go live.

LMK if you would like assistance updating the list of members, I can tackle that in a separate PR if you like.

TikhonJelvis:
Thanks. That'll be a tiny PR and I figure whoever ends up merging this one can do it right after—I'm just holding off for now because I don't want to think about merge conflicts when I should be going to bed instead :stuck_out_tongue:.

rebeccaskinner (approved):
This is a great start. I think we have some additional things we'll want to get added here, and hopefully this will kick off some momentum in us doing so.  Thanks for the PR!

ketzacoatl:
Great news, thanks for the collaboration and support along the way  :)

ketzacoatl:
https://wiki.haskell.org/Haskell.org_committee has been updated to drop the content and refer to https://www.haskell.org/haskell-org-committee/ and this PR.

SAGARGAUD01:
Hello sir how can I participate in this.

PR #69, opened 2021-03-07
Fix buildAndWatch command in nix-shell

TikhonJelvis:
In #63 we renamed the executable from `site` to `haskell-org-site` to avoid collisions with other projects. However, we didn't update the name in the `default.nix` file, which broke the `buildAndWatch` command in the Nix shell.

This PR fixes that problem.


ketzacoatl:
Sorry about that!

rebeccaskinner (approved):
LGTM

Issue #68, opened 2021-03-04
amazon smile link in the donations page broken


vu3rdd:
The amazon smile link on the [donations page](https://www.haskell.org/donations/) has not been populated with the [right link](https://smile.amazon.com/). Perhaps a little text explaining how to use that service and find haskell.org as a charity can help as well.

TikhonJelvis:
Thanks for spotting this. Just submitted a PR with a quick fix.

ketzacoatl:
@TikhonJelvis if that's been merged, is it good to close out this issue, or is there another task?

PR #70, opened 2021-03-07
Fixed link to Amazon Smile and added a more info link.

TikhonJelvis:
This should address #68.

I also fixed the link from the donations page to the committee page.


rebeccaskinner (approved):
LGTM

PR #75, opened 2021-03-27
site: find .markdown files 1, 2 or 3 levels down

ketzacoatl:
This commit lets the `haskell-org-site` match `.markdown` files several directory levels deep. This will support efforts to build out docs and tutorials, and as content is moved from the wiki.

### DIFF
--- a/site.hs
+++ b/site.hs
@@ -26,7 +26,7 @@ main = mkContext >>= \ctx -> hakyll $ do
     route idRoute
     compile $ defCompiler ctx
 
-  match ("**/*.markdown" .||. "*.markdown") $ do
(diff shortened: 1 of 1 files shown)


TikhonJelvis:
Hey, sorry, forgot to reply to this earlier. I'm pretty sure `**` patterns in Hakyll automatically work with any number of nested directories, but I haven't had a chance to try that yet—I'll confirm once I test that.

TikhonJelvis:
Yep, just had a chance to test this, and the `**` patterns definite match multiple levels of nesting. I believe we needed the `.||. "*.markdown"` bit because `**` matches *at least one* element, but this means that we don't need to explicitly cover the `**/**/*.markdown` case.

ketzacoatl:
Great, thanks!

PR #78, opened 2021-05-05
Update haskell.org committee members

tomjaguarpaw:
Tom Ellis replaces Emily Pillmore after she left to become Haskell
Foundation CTO.


TikhonJelvis (approved):
Looks good to me.

PR #80, opened 2021-05-06
docs: separate list of "additional resources" to a new page

ketzacoatl:
* Instead of the "documentation" page being a giant list of community resources, we can separate that as a list of resources the documentation page can link to.
* This opens the documentation page to host links to local docs, on the haskell.org site.
* For now, the list of links on the documentation page includes references to each of the existing/primary docs on the haskell.org site (donations, committee, downloads, community, learning resources).
* Pages can be moved from the wiki in the future, and added to this list of docs.
* Previously, the committee page was difficult to find a link to (only some pages have it), now this page makes that easier to locate.


ketzacoatl:
It looks like:

image

tomjaguarpaw:
I think the documentation page is rather bare now. Indeed the proposed "resources" page is basically exactly the same as the current "documentation" page. I agree with the rationale that

> This opens the documentation page to host links to local docs, on the haskell.org site.

but I think it might be worth waiting until those links to local docs actually exist, rather than anticipating something that has not yet happened.

ketzacoatl:
It is a bit of a catch-22, but sure, I can add new pages with more links on this documentation page.

tomjaguarpaw:
Cool, thank you.

ketzacoatl:
@tomjaguarpaw I took a crack at your suggested update, but running into issues.

The main one is that it's hard to make any meaningful change that would be widely accepted. I believe that is mostly a result of requirements either undefined, or poorly defined, or incomplete. We don't seem to define enough of the plan or requirements ahead of time, and so an implementer has to repeatedly battle with "is this even worth my time".

I started by adding the haskell.org links in a new section at the top of the page, put all other existing links under a separate h2, and then pulled a few haskell.org links from below up to the top. Stopping there, I could hear the various objections that would likely come, that the sections were now less clear, that the new section at the top was confusing and conflicting with the rest. I would mostly agree, but also need to communicate why this update was more of a stepping-stone, as well as what steps would come next.. but there are no clearly defined acceptance criteria or requirements that I as an implementer can lean on or refer to. 

When I started contributing doc changes to this repo at the beginning of the year, I gave myself a time limit and challenged myself to see what sorts of improvements or quantity of change I could reasonably muster. I'm pretty disappointed with my results, and will need to invest my time elsewhere for now. I would like to continue contributing to haskell.org in the future, but would need to see positive structural changes in haskell.org to feel that is worth my time.

Thank you for the interactions we've had and support you've provided @tomjaguarpaw, I look forward to working together in the future and around the community.

tomjaguarpaw:
Thanks for your contributions so far and hope to have you back again.

PR #82, opened 2021-05-07
Create a Zip File with GitHub Actions

TikhonJelvis:
This PR cleans up our GitHub action definition and generates a zip file with the site contents as an artifact.

The upload-artifact@v2 action had trouble with symlinks (see [this issue][1]), so we needed to specify the path as a glob.

Next steps: automatically deploy the site if it builds and passes our checks. This will require managing SSH keys in CI.

[1]: https://github.com/actions/upload-artifact/issues/92

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,3 +22,8 @@ jobs:
 
       - name: Check Links
         run: nix-shell --command 'linkchecker _site'
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: www.haskell.org
+          path: result-1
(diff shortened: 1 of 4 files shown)


tomjaguarpaw (approved):

TikhonJelvis:
Another thing that could be worth doing:

  1. Refactor executable and site content into separate directories
  2. Use Cachix for caching intermediate results

This should speed up CI and development times by not rebuilding the executable unnecessarily and caching our artifacts between builds.

tomjaguarpaw:
As per https://github.com/haskell-org/committee/blob/main/proposals/0003-pr-process.md I shall merge this.

PR #85, opened 2021-05-25
Remove broken webmon.com link

djanatyn:
It appears this domain is no longer pointing towards a legitimate service, at least as of Feb 13th 2021:
* https://www.domainiq.com/domain?webmon.com

It might be best to remove this link from the https://www.haskell.org, at least for now.

Currently, the domain is parked on both 443 and 80:

webmon1
webmon2

### DIFF
--- a/index.html
+++ b/index.html
@@ -468,11 +468,6 @@ isHome: true
             <p><strong><a href="https://www.dreamhost.com">DreamHost</a></strong> has teamed up to provide Haskell.org with redundant, scalable object-storage through their Dream Objects service.</p>
          </div>
       </div>
-      <div class=" row ">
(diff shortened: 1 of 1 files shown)


tomjaguarpaw (approved):
Thank you for spotting this and submitting the PR!

tomjaguarpaw:
Looks like now that we have https://github.com/haskell-org/committee/blob/main/proposals/0003-pr-process.md I can merge and deploy so will do so.

tomjaguarpaw:
Sorry, I'm having trouble deploying this. I will seek advice from the rest of the committee.

tomjaguarpaw:
I take that back. It *has* deployed.

PR #86, opened 2021-05-26
Fix haskell language channel names for new network

davean:
The language channels are wrong for the new network. Its causing confusion.


tomjaguarpaw (approved):
Thank you for fixing

tomjaguarpaw:
Looks like this falls under item 2 of https://github.com/haskell-org/committee/blob/main/proposals/0003-pr-process.md, so I will merge and deploy immediately.

tomjaguarpaw:
Looks like someone had already deployed this, actually.

Issue #83, opened 2021-05-12
Set up automated deployment from CI


tomjaguarpaw:
Davean explained to me how we can automate deployments from CI by via sftp. The process is

* store an SSH private key as [a secret in the CI system](https://docs.github.com/en/actions/reference/encrypted-secrets)
* send the public key to `admin@haskell.org` for the admins to install on the server
* each CI build of master, `sftp` the site artifact directly to the `www.haskell.org` server

cc @TikhonJelvis 

Related to https://github.com/haskell-infra/www.haskell.org/issues/66

tomjaguarpaw:
First step would be to build the artifact. This is ready as https://github.com/haskell-infra/www.haskell.org/pull/82.

ketzacoatl:
This is exciting! While I believe a simpler setup with static hosting would be better suited to the situation, I am glad to see CI improving!

PR #87, opened 2021-05-26
Automatically deploy site

tomjaguarpaw:
This PR adds a Github Action to build the website and deploy it to the haskell.org server.  We deploy the site *only* on

1. A push to `master`, or
2. A manually triggered Deploy build

(We don't want to deploy for every branch and PR of course!)

I have tested this on my fork. See https://github.com/tomjaguarpaw/www.haskell.org/runs/2678438965?check_suite_focus=true for a successful (dry run) deployment.

### What we need from the Haskell infrastructure admins

- [x] Confirmation that this matches their expectations of an automated deployment script
- [x] A definitive username and directory on the remote host to be the target of the `rsync`
- [x] To authorise the `rsync` user by placing the public key on the server in the right place

### To do before merging:

- [x] Generate an SSH public/private key, get the public key on the server via @davean/Haskell infrastructure admins, get the private key into Github Actions secrets
- [x] Convince ourselves that it is safe to remove `--dry-run`

(Closes https://github.com/haskell-infra/www.haskell.org/issues/83)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,51 @@
+name: Deploy
+
+# We deploy the site *only* on
+#
+# 1. A push to master, or
+# 2. A manually triggered Deploy build
+#
+# We don't want to deploy for every branch and PR of course!
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v12
+
+      # The CI Github Action also builds the site and puts it in an
+      # artifact.  However, downloading an artifact from a different
+      # workflow seems to be difficult.
+      #
+      # https://github.com/actions/download-artifact/issues/3
+      #
+      # We'll just build the whole site again here.  It only takes a
+      # couple of minutes.
+      - name: Build Site
+        # we run the checks as a separate step
+        run: nix build -f . built -o built-site --arg doCheck false
+
+      - name: Check Links
+        run: nix run --quiet -f . linkchecker -c linkchecker built-site
+
+      - name: Deploy
+        shell: bash
+        env:
+          # Create the secret with:
+          #
+          # Settings -> Secrets -> New Repository Secret ->
+          #
+          #   Name: WWW_HASKELL_ORG_SSH_KEY
+          #   Value: The text of the ssh private key
+          WWW_HASKELL_ORG_SSH_KEY: ${{ secrets.WWW_HASKELL_ORG_SSH_KEY }}
+        run: |
+          nix run --quiet -f . lftp -c sh ./deploy.sh --DEPLOY-IT-LIVE
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,0 +1,21 @@
+#!/bin/sh
+
+# TODO: make this with tempfile
+KEYFILE=__TEMP_INPUT_KEY_FILE
+
+echo "${WWW_HASKELL_ORG_SSH_KEY}" > "$KEYFILE"
+chmod 600 "$KEYFILE"
+
+if [ "$1" = --DEPLOY-IT-LIVE ]; then
+    DESTINATION=www
+else
+    DESTINATION=www-test
+fi
+
+# WARNING: --delete is dangerous.  Be absolutely sure you are
+# deploying to the correct directory.  Existing contents of the
+# destination directory will be removed.
+#
+# Permissions: We have to recursively set write permissions because
+# lftp isn't able to remove a file if its directory is not writeable.
+lftp --norc -c "set net:max-retries 1; set sftp:connect-program \"ssh -o StrictHostKeyChecking=no -a -x -i $KEYFILE\"; open -u www-sftp, sftp://webhost.haskell.org; chmod --recursive u+w $DESTINATION; mirror --delete --verbose --reverse --transfer-all built-site $DESTINATION"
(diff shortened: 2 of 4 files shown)


TikhonJelvis (approved):
Looks great to me. Once @davean is happy, we can merge this and have low-friction deploys :)

tomjaguarpaw:
Pinging @davean

tomjaguarpaw:
I have asked @davean about this on #haskell-infrastructure

tomjaguarpaw:
Perhaps @davean is busy at the moment. I have pinged @gbaz on #haskell-infrastructure. Perhaps he can help us instead.

davean:
The host name should be "webhost.haskell.org" for reliability - www-combo-nix.haskell.org is the specific name of this system not the alias in general.

While rsync works with the user account, but for better security
`lftp --norc -c	"open -u www-sftp,ssh-key sftp://webhost.haskell.org; mirror -Re --transfer-all to_upload www"`
works with a locked down sftp-only user.

I tried to come up with a better version of this failed. The problems I identified specifically all relate to write tearing. If multiple, different, uploads are attempted at once, or a partial upload leave issues with the site. Sadly rename(2) can't work with non-empty directories, and while a link would be atomic I couldn't figure out how to clean up old versions safely and atomically.

For testing, I've setup the sftp limited version as a test site at www-test.haskell.org.

tomjaguarpaw:
Thanks @davean! I will get back to this as soon as I can.

tomjaguarpaw:
I made some minor changes to the suggested `lftp` command `lftp --norc -c "set net:max-retries 1; set sftp:connect-program \"ssh -o StrictHostKeyChecking=no -a -x -i $KEYFILE\"; open -u www-sftp, sftp://webhost.haskell.org; mirror --delete --verbose --reverse --transfer-all built-site www"`

tomjaguarpaw:
This is ready to be merged. I shall ask the committee members to have a look since is a significant change to how the site will be deployed.

deploy.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+# TODO: make this with tempfile
+KEYFILE=__TEMP_INPUT_KEY_FILE
+
+echo "${WWW_HASKELL_ORG_SSH_KEY}" > "$KEYFILE"
+chmod 600 "$KEYFILE"
+
+# WARNING: --delete is dangerous.  Be absolutely sure you are
```
TikhonJelvis, line 9: Would it make sense to do a dry run by default, and have a special `--run` flag in this script to do the actual `--delete`? The CI script could pass the flag in, but this way the default isn't destructive if anyone is testing things manually.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # Should probably do this with Nix
+      - name: Install lftp
+        run: sudo apt-get install lftp
```
TikhonJelvis, line 24: Yep, we can do this with Nix. To use the same patter as we did for `linkchecker`—seems reasonable—we would need to add `lftp` as an attribute to the output of `default.nix` then use the following command:

```
nix run -f . lftp -c lftp <lftp options>
```

We could also pass the `--quiet` flag to `nix run` if we want to suppress Nix's own output:

```
nix run --quiet -f . lftp -c lftp <...>
```

The `.` refers to the `default.nix` file at the top level of the repo; to run from a different directory, we would need to change that path.

I haven't tested this on GitHub Actions, but it worked locally:

```
>nix run -f . lftp -c lftp --help
Usage: lftp [OPTS] <site>
`lftp' is the first command executed by lftp after rc files
 -f <file>           execute commands from the file and exit
...
```

I'll push a commit with these changes; if it doesn't work, we can always revert it :)

TikhonJelvis: Actually, not sure if I can just push a commit to ` tomjaguarpaw:master`. If that doesn't work, I might just leave the changes as GitHub suggestions since they're pretty small.

TikhonJelvis: Down the line, we'll probably want to organize this a bit better—something like a `ci.nix` file that has all the tools and dependencies we need for CI, so that we can manage everything from one place and don't have to use `nix run` all the time. For now I'll go with the quicker solution, but this is something we should think of refactoring, maybe the next time we want to add a new tool or dependency to CI.

TikhonJelvis: Okay, see [this PR](https://github.com/tomjaguarpaw/www.haskell.org/pull/1) for the changes :)

romefeller (approved):

.github/workflows/deploy.yml
tomjaguarpaw: I incorporated your change to get `lftp` via nix, thanks!

deploy.sh
tomjaguarpaw: I think Davean will set us up a `www-test` directory that will be deployed to https://www-test.haskell.org/. It would be good to have a flag on this script which chooses between live and test. I will confirm with Davean.

tomjaguarpaw: OK, looks like this is in the pipeline so I added that flag. Not yet enabled.

tomjaguarpaw:
This is ready for merge. Anyone else want to have a look? @jaspervdj @rebeccaskinner @idabzo @ryantrinkle

deploy.sh
tomjaguarpaw: Now enabled

tomjaguarpaw:
Thanks to everyone who has commented so far. I'll leave this open with the aim of merging next weekend if there is no dissent.

rebeccaskinner (approved):
Looks good, I think this will be a great thing to have.

tomjaguarpaw:
I have merged but it seems there is an authorisation problem. I will sort it out with davean.

PR #88, opened 2021-05-27
Adding a getting started tutorial describing the core GHC tools

soupi:
@Kleidukos

This PR adds a tutorial describing the core GHC tools. It covers common usage of `ghc`, `runghc` and `ghci` as well as slightly more advanced usage, and linking to the GHC manual where appropriate.

This page can be used as a first step towards a getting started guide and can be linked from the documentation or downloads pages.

Similar ideas have been previously requested [here](https://github.com/haskell-infra/www.haskell.org/pull/71#issuecomment-827812112).

[Rendered](https://soupi.github.io/www.haskell.org/core-tools/index.html)


tomjaguarpaw:
Thanks! In case you don't know you can find a built artifact at https://github.com/haskell-infra/www.haskell.org/actions/workflows/main.yml which will show everyone what your changes look like.

Kleidukos (approved):

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
+---
+title: Learn about the core tools
+page: core-tools
+isDocumentation: true
+---
+
+# Learn about the core tools
```
rebeccaskinner, line 7: Given the audience for this article, I wonder if we wouldn't be better to actually pick an installation method and walk through it to start with (or we can have that as another article and link back to it). I know that picking a single installation method will be controversial, but it seems like this document is suffering a lot from the lack of being able to assume a specific configuration.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+
+In this guide we'll take a look at a few core tools that are installed
+with the Haskell toolchain, namely, `ghc`, `runghc` and `ghci`.
+These tools can be used to compile, interpret or explore Haskell programs.
+
+Note: if you installed your Haskell toolchain using the stack tool only
```
rebeccaskinner, line 13: See above, while I think this note is perfectly reasonable if we're not making assumptions about how things are installed, I think it would be a much better document if we just assumed an installation method.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+
+```sh
+➜ ghc hello.hs
+[1 of 1] Compiling Main             ( hello.hs, hello.o )
+Linking hello ...
+```
+
+GHC created the following files:
+
+1. `hello.hi` - Haskell interface file, we won't discuss that in this article
```
rebeccaskinner, line 62: If we're not going to talk about the hi and o files, maybe we shouldn't bring them up here? Perhaps something like "When we compile the application, GHC will dutifully creating our new executable, `hello` (or `hello.exe` on windows). You may notice a couple of other files that get generated too, but we won't worry about them in this tutorial."

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+2. `hello.o` - Object file, the output of the compiler before linking, we won't discuss that in this article either.
+3. `hello` (or `hello.exe` on Microsoft Windows) - A native runnable executable.
+
+GHC will produce an executable when the source file:
+
+1. Defines the binding `main` in the source file
```
rebeccaskinner, line 68: Maybe we should just call `main` a function for the sake of familiarity here?

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+2. Defines the module name to be `Main` (this can be done by adding `module Main where` at the top of the file), or does not have a module declaration (which is then inferred as the module `Main`).
```
rebeccaskinner, line 69: I wonder if we might not want to include the module in initial hello world example too, rather than mentioning it here for the first time.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+To exit `ghci`, use the `:quit` command (or `:q` for short)
+
+```sh
+ghci> :quit
+Leaving GHCi.
+```
+
+A more thorough introduction to GHCi can be found in the [GHC user guide](https://downloads.haskell.org/~ghc/9.0.1/docs/html/users_guide/ghci.html).
+
+### Using external packages in ghci
```
rebeccaskinner, line 227: Again I wonder if we should maybe just presume one of cabal or stack and stick with it for the tutorial rather than trying to show it both ways.

rebeccaskinner (changes requested):
Thanks for the PR! Having more documentation like this will be incredibly valuable.

I think there are a few opportunities for cleanup here that could make this a lot more accessible. I made some specific notes in the PR, but more broadly:

I don't see where this is linked from the rest of the site- do we want to add a link to the top-level documentation page to this so users can find it, or will we open up this later along with other articles?

I am having a hard time figuring out who the audience is for this article. Early on it addresses a reader who has seemingly never used haskell before at all, but at the end of the article it starts talking about RTS options and profiling. I don't think the same people are going to benefit from both of those things at the same time, so I think we'd be better off breaking that up.

soupi (commented):
Thanks for the review!

> I don't see where this is linked from the rest of the site- do we want to add a link to the top-level documentation page to this so users can find it, or will we open up this later along with other articles?

I talked to @Kleidukos about this as well and I am not sure where this document should be linked, the ideas we had are from the download page,  the documentation page (thought it is in a dire need of revamp), a new getting started page that could be made or even school.haskell.org. I don't know what is the right place and would like to open this for discussion.

> I am having a hard time figuring out who the audience is for this article. Early on it addresses a reader who has seemingly never used haskell before at all, but at the end of the article it starts talking about RTS options and profiling. I don't think the same people are going to benefit from both of those things at the same time, so I think we'd be better off breaking that up.

The audience are indeed readers who have never used Haskell before, but I think it is a good opportunity to also cater to a larger portion of Haskell users by including a slightly more advanced (but still quite common) usage section that unfortunately (to my eyes, I may be mistaken) just isn't very well known/covered by other popular material, so I wanted to "save the trouble" of looking it up .

Basically what I wanted to say is "hey new user, when you are ready, you can find information about these very common things here". This might not be the best way to do it and I can remove this section if you think it's better.

Perhaps a simple sentence such as "The GHC manual contains more useful information on how to control the Haskell runtime and how to profile Haskell programs" would suffice.

core-tools.markdown
Kleidukos: Yes, call it a function, this is the term that will be expected of us

core-tools.markdown
Kleidukos: We can use :

```shell
$ ghc -e 'putStrLn "hello world!"'  
hello world!
```
for a one-liner

core-tools.markdown
Kleidukos: Agreed, this is the opportunity to unify an approach to both Cabal & Stack.

Kleidukos:
Regarding the RTS page, maybe a one-page lesson on the Haskell School would do the job better than a one-paragraph overview?

core-tools.markdown
soupi: I don't think one-liner is the goal though :) should we even cover `-e` at all? I've never used it myself. I would rather just pick whether to use a module definition or not.

core-tools.markdown
soupi, line 68:
```suggestion
1. Defines the main function in the source file
```

soupi:
> Regarding the RTS page, maybe a one-page lesson on the Haskell School would do the job better than a one-paragraph overview?

core-tools.markdown
ulysses4ever: My 5c: it is really annoying to have to jump over every other sentence in a section because it mentions "the other" tool that I don't use. Please, don't do that. I really hope that this page, if it covers both tools, does this in **separate (sub)sections** even if these sections are small.

soupi: I don't think you'll experience such annoyances in this article because this happens exactly once in the article in an extremely tiny section. I'm open to seeing an alternative presentation of the information though.

core-tools.markdown
rebeccaskinner: The main thing is that I think the prose sometimes becomes a bit hard to follow because of the need to talk around multiple setups. Being able to say: "Type `xyz` into your terminal" is a powerful thing in terms of making the tutorial easy for people to follow. As a secondary issue, I think if we're aiming this at early beginners, the fact that there are many choices is going to be overwhelming to them.

soupi: I think I could understand better if you could please mention places where you think the prose becomes a bit hard to follow and could be improved.

I don't quite understand because except for one instance (using packages) this tutorial does not talk about multiple setups. And it does not asks the users to make choices.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+
+By default, ghci can only load and use packages that are [included with the GHC installation](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/9.0.1-notes.html#included-libraries).
+
+However, users of [cabal-install](https://www.haskell.org/cabal) and [stack](http://haskellstack.org) can download and load external packages very easily using the following commands:
+
+Cabal-install:
+
+```sh
+cabal repl --build-depends random
```
ulysses4ever, line 236: I suggest adding or modifying the example to show how to load more than one package because the syntax for that is not immediately obvious from that for one package. (Both cabal and stack.)

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+  Alloc rate    717,569,877 bytes per MUT second
+
+  Productivity  27.4% of total user, 28.0% of total elapsed
+```
+
+This information is not very useful for such a trivial program,
+but as programs get more complex it will provide more insight
+regarding the runtime of the program.
+
+Alternatively, it is also possible to set the exact runtime flags
```
ulysses4ever, line 315: I'd skip this fact entirely from this kind of first intro to the tooling. It takes bunch of space in the document but hardly brings much profit.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+➜ ghc -rtsopts hello.hs 
+[1 of 1] Compiling Main             ( hello.hs, hello.o )
+Linking hello ...
+```
+
+And run it with `-sstderr` which will print some information
+about the runtime execution of the program: 
+
+```sh
+➜  ./hello +RTS -sstderr -RTS
```
ulysses4ever, line 286: This is not a very compelling case for `-rtsopts` because you can get the same info with `./hello +RTS -s` without `-rtsopts` during compilation.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+λ> import System.Random
+λ> randomRIO (1, 10)
+1
+λ> randomRIO (1, 10)
+8
+λ> randomRIO (1, 10)
+5
+```
+
+Stack users can also use this feature with `runghc` and `ghc` by replacing `ghci` in the command above.
```
ulysses4ever, line 258: For symmetry I'd add that `cabal --lib --package-env . install random` will generate an environment file that will make `random` visible for GHC / runghc calls in the current directory. Alternatively, I'd remove this topic altogether.

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+
+Note that `-Wall` does not stop GHC from compiling your program.
+If this is something you'd like to do, specifying the flag `-Werror`
+will convert warnings to errors and will halt compilation when a warning is emitted.
+
+#### Optimisations
+
+Another very useful flag is `-O`. This flag asks GHC to compile the program with optimisations and code improvements at the cost of taking longer to compile.
+This will often make the program run much faster and perform more aggressive optimisations such as inlining and specialisation.
+For even more optimisations that may take significantly longer, `-O2` is also available.
```
ulysses4ever, line 157:
```suggestion
For even more optimisations that may take significantly longer to compile, `-O2` is also available.
```
Some redundancy will make it clearer.

core-tools.markdown
ulysses4ever: It doesn't happen exactly once unfortunately: lines 13, 231, 258 all introduce tool-related topics. After reading the whole document I don't have better suggestions though, so I'd leave it as it is probably but add the cabal-related solution to the issue raised on line 258 (see my comment there).

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+Otherwise, it will only produce the `.o` and `.hi` files.
+
+In our case, we have defined `main` and omitted the module declaration, so GHC created an executable for us. And we can run it:
+
+```sh
+➜ ./hello 
+Hello, Haskell!
+```
+
+Or on Microsoft Windows:
```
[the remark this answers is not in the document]
soupi: Good to know! Thanks

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+```
+
+`runghc` interprets the source file instead of compiling it and does not
+create build artifacts. This makes it very useful when developing programs
+and can help accelerate the feedback loop. More information about `runghc`
+can be found in the [GHC user guide](https://downloads.haskell.org/ghc/latest/docs/html/users_guide/runghc.html).
+
+### Common options for GHC
+
+Here are a few notable options we can use with `ghc` and `runghc`.
```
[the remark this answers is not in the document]
soupi: Good idea

core-tools.markdown
```diff
@@ -0,0 +1,345 @@
...
+`-Wall` emits many useful warnings that can easily be bugs,
+including warnings about name shadowing, unused variables, and more.
+
+Note that `-Wall` does not stop GHC from compiling your program.
+If this is something you'd like to do, specifying the flag `-Werror`
+will convert warnings to errors and will halt compilation when a warning is emitted.
+
+#### Optimisations
+
+Another very useful flag is `-O`. This flag asks GHC to compile the program with optimisations and code improvements at the cost of taking longer to compile.
```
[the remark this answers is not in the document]
soupi: There isn't, but I'll figure something out.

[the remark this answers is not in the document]
soupi: I can work with whatever convention is required

core-tools.markdown
soupi: You are right. I think according to feedback on this section it is better to remove the section altogether and link the GHC manual for more details.

core-tools.markdown
soupi: I'd very much like to add such a command to the tutorial, but I tried running said command but ran into an error:

```sh
➜  cabal --lib --package-env . install random
cabal: unrecognised command: . (try --help)
```

soupi:
I've applied the github suggestions received, removed the windows specific usage, replaced the external packages example with one that uses two packages instead of one, and replaced the advanced usage section with a link to the ghc manual for additional information.

Updated tutorial render is available [in the same link](https://soupi.github.io/www.haskell.org/core-tools/index.html).

core-tools.markdown
ulysses4ever: I'm sorry, should have checked: `install` should go first:

```
cabal install --lib random --package-env .
```

core-tools.markdown
```diff
@@ -0,0 +1,301 @@
...
+stack exec --package async --package say -- ghci
+```
+
+And the modules of the relevant packages will be available for import:
+
+```sh
+GHCi, version 9.0.1: https://www.haskell.org/ghc/  :? for help
+ghci> import Control.Concurrent.Async 
+ghci> import Say
+ghci> concurrently_ (say "Hello") (say "World")
```
soupi, line 277:
```suggestion
ghci> concurrently_ (sayString "Hello") (sayString "World")
```

core-tools.markdown
```diff
@@ -0,0 +1,301 @@
...
+Hello
+World
+```
+
+Stack users can also use this feature with `runghc` and `ghc` by replacing
+`ghci` in the command above. <!-- , and Cabal-install users can generate an
```
soupi, line 283:
```suggestion
`ghci` in the command above, and Cabal-install users can generate an
```

core-tools.markdown
```diff
@@ -0,0 +1,301 @@
...
+environment file that will make `async` and `say` visible for GHC tools
+in the current directory using this command:
+
+```sh
+cabal --lib --package-env . install async say
```
soupi, line 288:
```suggestion
cabal install --lib async --lib say --package-env .
```

core-tools.markdown
```diff
@@ -0,0 +1,301 @@
...
+```
+-->
```
soupi, line 290:
```suggestion
```

core-tools.markdown
ulysses4ever: I don’t know: I think `-s` is pretty cool (in terms of all the info that it gives you for free). And I’m not big fun of sections like the last one on profiling where there’s just a reference and that’s it — it feels underwhelming. So it’s all about balance I guess: too much is overwhelming, too little is vice versa... It’s s your call in the end!

soupi: I also think it's cool :) but I think in accordance with feedback from Rebecca and others it might be best to leave this kind of things out for now. We could always revisit this later and add it if we wish.

core-tools.markdown
soupi: Thanks! I've added this option and replaced the example with an example that requires two packages.

core-tools.markdown
```diff
@@ -0,0 +1,298 @@
...
+GHCi, version 9.0.1: https://www.haskell.org/ghc/  :? for help
+ghci> import Control.Concurrent.Async 
+ghci> import Say
+ghci> concurrently_ (sayString "Hello") (sayString "World")
+Hello
+World
+```
+
+Stack users can also use this feature with `runghc` and `ghc` by replacing
+`ghci` in the command above, and Cabal-install users can generate an
```
ulysses4ever, line 281: "Cabal-install" should really be "cabal-install" (here and elsewhere). It's only Cabal the library that is capitalized.

core-tools.markdown
```diff
@@ -0,0 +1,298 @@
...
+environment file that will make `async` and `say` visible for GHC tools
+in the current directory using this command:
+
+```sh
+cabal install --lib async --lib say --package-env .
```
ulysses4ever, line 286: As far as I can see, one "--lib" is enough.

soupi:
Where should the link to this document go? A few ideas are:
- the downloads page
- the documentation page
- A new getting started page (that doesn't exist yet)

core-tools.markdown
hasufell: @rebeccaskinner I believe installation instructions have no place here, that's what https://www.haskell.org/downloads/ is for

hasufell:
Yes, I think this should be linked from the downloads page... however... for me it's missing a big section (which is currently not covered by the downloads page either), which is: explaining the difference of cabal and stack. And possible giving guidance as to which tool to choose.

That's the main thing missing from the downloads page (and it shouldn't be there).

Maybe it's out of scope of **this** PR, but we do need it.

soupi:
I think the issues we still need a resolution for in this PR are:

1. Should we include the note about using stack as an installation method or not? ([link](https://github.com/haskell-infra/www.haskell.org/pull/88/files#diff-6e06515857e7603e2d734c3a97cde77cf7d00344596c99c8d257e98e5526b49aR13))
2. Should we mention the aux files created by ghc? ([link](https://github.com/haskell-infra/www.haskell.org/pull/88/files#diff-6e06515857e7603e2d734c3a97cde77cf7d00344596c99c8d257e98e5526b49aR62))
3. Are we okay with the external packages section? ([link](https://github.com/haskell-infra/www.haskell.org/pull/88/files#diff-6e06515857e7603e2d734c3a97cde77cf7d00344596c99c8d257e98e5526b49aR248))
4. Where should we place the link to this document?

Is there anything I missed?

tomjaguarpaw:
Sorry I haven't been able to look at this yet. I will do so as soon as I get the opportunity.

core-tools.markdown
```diff
@@ -0,0 +1,298 @@
...
+Many more packages are waiting for you on [Hackage](https://hackage.haskell.org).
+
+## Additional information
+
+This article covered the most common usage of the core tools GHC offers.
+
+If you'd like to learn more about them,
+[the GHC manual](This article describes the most common usage of the core tools)
+Contains additional information on how to use the tools, including how to control
+the runtime system and how to profile Haskell programs.
```
tomjaguarpaw, line 298: Something funny is going on here

Kleidukos: More precisely, the markdown notation is inverted

rebeccaskinner:
I think the content is looking better here with the edits. There are two major questions that are still standing out in my mind:

The first is the more immediate question of who our target reader is for this. It seems to be written for a new user who is just getting started with haskell, but I'm having a hard time picturing a scenario where a user who is just getting started is going to interact with ghc this directly. I'm open to the idea that I might be wrong here, but it seems to me that most users are going to install stack or cabal and just do everything through there- and in general I'd think that's probably what we'd want to encourage for the most part.  I hope this isn't taken to mean that this or similar documentation aren't needed- I think it's really good that we're starting to look at this, I just want to be very mindful of ensuring that we have a very clear set of users in mind with the docs that we write, and understand how the docs help them on their journey to learn and use haskell effectively.

The second question is a bit less immediate, but I'd like to understand where we're going with this stuff in general. Is this just one additional document that we want to have to supplement the information on the downloads page, or is it part of some larger work that we need to consider? In either case I think we need to figure out how we want to link to it and where it's going to live.

gbaz:
In my experience, a lot of new users want to just learn the language, which means interacting with ghci, and compiling simple example executables, without starting a project or introducing any external deps at all. This especially holds in many classroom settings.

TikhonJelvis:
When I started, I spent some time just using a REPL and then started using `cabal`. I never tried building executables with `ghc` directly and it has almost never come up in a decade of active Haskell use. Going from basic Haskell in GHCi right to a Cabal/Stack project seems like an even more natural path today since both tools are substantial improvements over 2010-era Cabal.

soupi:
@rebeccaskinner this is an interesting point. In my opinion for anything that has less than a few package dependencies, using `ghc`, `runghc` and `ghci` is much easier than creating packages, and I always suggests starting off with these tools in my beginner tutorials instead of trying to figure out cabal or stack (and which one to use) so learners can focus on Haskell rather than it's tooling.

For example, a learner that wants to create a guessing game can just use

```
stack exec --package random -- runghc my_guessing_game.hs
```
or similar to run their program without having to figure out anything else. In my opinion this lowers the barrier of entry.

But I also understand your opinion saying that using stack/cabal is inevitable for larger programs so this article might not be useful. This is really up to you to decide whether this is the right content to have on haskell.org and whatever you decide is fine with me.

> The second question is a bit less immediate, but I'd like to understand where we're going with this stuff in general. Is this just one additional document that we want to have to supplement the information on the downloads page, or is it part of some larger work that we need to consider?

I definitely support this conversation. cc @Kleidukos

core-tools.markdown
soupi: whoops. thank you - should be fixed now.

tomjaguarpaw:
I think it would be good to have a meeting between the Haskell.org committee and the documentation task force to brainstorm the best way for Haskell.org to host great documentation for Haskell!

Kleidukos:
On my end, I have become a stakeholder in this effort because the first lesson of Haskell School will require the reader to have a basic understand of how to use ghci/cabal/stack for *simple* snippets of code (that may depend on external dependencies like containers or text).

I definitely agree that a call is warranted so that we can synchronise our clocks. @tomjaguarpaw Can I let you come back to me once you know who from the h.org committee would like to participate in the call?

tomjaguarpaw:
OK, I will take care of arranging that on the Haskell.org end.

soupi:
ping

core-tools.markdown
```diff
@@ -0,0 +1,297 @@
...
+World
+```
+
+Stack users can also use this feature with `runghc` and `ghc` by replacing
+`ghci` in the command above, and cabal-install users can generate an
+environment file that will make `async` and `say` visible for GHC tools
+in the current directory using this command:
+
+```sh
+cabal install --lib async say --package-env .
```
hasufell, line 286: `cabal install --lib` adds stuff to the default environment, which can cause very confusing build errors later when building projects. I'd advise to not mention it at all.

Instead, use `cabal repl --build-depends async`.

soupi: done. thanks.

ulysses4ever: @hasufell that's not totally correct: the command as was written worked just fine and didn't touch the global environment. That's because the `--package-env .` part. I'd vote for putting it back. But up to you guys.

hasufell: @ulysses4ever Thanks, I overlooked that.

soupi: I see you both agree that this is actually ok, so I'll revert the change.

ulysses4ever: I also double-checked that preceding text tries to explain what's happening in a nutshell. But maybe Julian has a point here and it would make sense to add a warning (after the command):

> (Warning: the command may work without the `--package-env .` part but in that case it will use the global environment instead of the local one. This is known to lead to complicated problems in some cases and discouraged.)

Of course, the warning itself may sound too confusing for a novice, as well as simply extending the text that we'd like to keep short, so I am not 100% sure.

soupi: I'd like avoid these kind of warnings in this text.  I also don't think it's relevant because the specific command we teach isn't problematic. If users want to experiment with different commands than what we show they will have to figure it out from other texts.

ulysses4ever:
@soupi would it be possible to post a comment about why this was closed?

soupi:
> @soupi would it be possible to post a comment about why this was closed?

After several months of no response/indication of progress, this PR doesn't feel like it's going to get merged, so I'm closing it to clear it off my plate.

hasufell:
I'm not part of the haskell.org committee, so I cannot approve. Sorry.

ulysses4ever:
@soupi fair enough, thank you for the answer and for your work.

gbaz:
Well hopefully @Kleidukos can get this merged into the school of haskell work?

Kleidukos:
I'd rather hear from @tomjaguarpaw since he said he would arrange it on the Haskell Committee end, before doing anything on my own without coordination from one of the stakeholders.

gbaz:
@Kleidukos I mean, this PR was just closed in this repo because the submitter gave up on getting it merged here. So I would hope the work can be adopted elsewhere if its not going here.

Kleidukos:
If @soupi allows it, I would like to petition the Haskell.org Committee to adopt this work and bring it to fruition.

tomjaguarpaw:
Sorry if you were waiting for me on this. I don't have any news but I'll bring it to the attention of the committee.

tomjaguarpaw:
After having raised this with committee, our decision is that we would love to link to the School of Haskell (including the  article in this PR) or perhaps indeed host the School of Haskell under haskell.org in due course. However, this one article feels like it would be out of place on haskell.org, without putting it in the additional context of the Haskell school.

Speaking personally, it seems to me that changes to haskell.org beyond the level of "minor" are slow to make for historical reasons, party to do with fear of reopening old controversies. I am optimistic that the situation will slowly improve over time. In the meantime I suggest that bold new web content start its life somewhere off haskell.org and we aim to bring it onto haskell.org once the community has had the opportunity to see its value.

hasufell:
> After having raised this with committee, our decision is that we would love to link to the School of Haskell (including the article in this PR) or perhaps indeed host the School of Haskell under haskell.org in due course. However, this one article feels like it would be out of place on haskell.org, without putting it in the additional context of the Haskell school.

I'm confused by this. [school of haskell](https://www.schoolofhaskell.com/) is discontinued and is read-only. https://school.haskell.org/ points nowhere and https://github.com/haskellfoundation/HaskellSchool gives no information about where it is hosted either.

If someone does host it, there seems to be an issue with discoverability.

tomjaguarpaw:
Perhaps @Kleidukos can provide an update of School of Haskell.

hasufell:
I've used this PR and added it to the ghcup homepage, simplifying some sections and adding another section about cabal project creation: https://www.haskell.org/ghcup/steps/

Kleidukos:
Hi, indeed Haskell School isn't being hosted anywhere. 
Other duties (both in the HF and elsewhere) have prevented me from dedicating the same amount of time to the project, although I expect to make more time this month.

PR #90, opened 2021-05-28
Add link to Haskell Matrix Space

srid:
The Matrix Space includes bridges to the IRC channel like #haskell as well.

Here's a screenshot of our "room tree".

image
image

### DIFF
--- a/community.markdown
+++ b/community.markdown
@@ -17,6 +17,7 @@ Haskellers are active on a number of online areas, but the most busy are below:
 
 *   [The Haskell mailing lists](/mailing-lists/)
 *   [IRC (online chat)](/irc/)
+*   [Matrix (online chat)](https://matrix.to/#/#haskell-space:matrix.org)
 *   [Haskell Community Discourse](https://discourse.haskell.org)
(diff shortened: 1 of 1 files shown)


simonmichael:
I think the link needs a bit more thought/explanation, since right now most matrix users will see a read only room.

rebeccaskinner:
Should we hold off on spaces until they are stable? I keep hitting bugs with the element web client making me sign back up for the beta every time I log in, and the current version of the element desktop client doesn't seem to support them at all yet. I'm not very up with Matrix overall though, so I'm not sure if everyone else is already on-board with these or not.

srid:
I've changed the link to not use spaces. It links directly to the room now. 

We can change it back to spaces, once Element sufficiently improves the spaces UX.

rebeccaskinner (approved):
Looks good to me- thanks for the contribution!

tomjaguarpaw (approved):
Thanks for this!

(As a side note, I understand the linked Matrix room has no active maintainer. If this is true then that should be addressed (and if necessary this link updated), although I'm not suggesting that stand in the way of providing the link here.)

tomjaguarpaw:
Thanks!

PR #91, opened 2021-05-29
Add #haskell-language-server channel

jneira:


rebeccaskinner (approved):
This looks good to me.

tomjaguarpaw (approved):
Thanks!

(In fact this probably belongs under "Related channels" rather than "Projects using haskell" but I don't mind merging this now and then fixing that up later. Lots of other things on that page need fixing too, like alphabetisation and splitting off the language-specific channels from the "Related channels".)

jneira:
agree, but I was not sure about so I will move to related channels, thanks!

jneira:
I am not used to editing using the web UI in a phone. 😟

jneira:
@tomjaguarpaw ok, i think it is ready to go, thanks!

tomjaguarpaw:
Thanks! There seems to be a problem with the deployment system so it may take a while before this reaches the live site. Feel free to comment on this PR (or contact me directly) if that becomes a problem.

PR #94, opened 2021-06-12
refactor buildAndWatch to run outside of the nix shell

rebeccaskinner:
This change is intended to make it a little bit easier for new contributors to get started, and to address an issue with the buildAndWatch shell alias inside of the nix shell.

Change summary:

- Remove the `buildAndWatch` shell alias from the builder's nix-shell, since this didn't seem to work properly
- Add a `buildAndWatch` shell script that provides the same functionality as the shell alias, but runs from outside of a nix shell
- Update the README file to point to the new `buildAndWatch` script, and to try to present an easier to follow workflow for making changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,25 +11,70 @@ Some of the others are
 * www.haskell.org/platform (built from [haskell-platform](https://github.com/haskell/haskell-platform/tree/master/website))
 * www.haskell.org/ghcup (build from [ghcup-hs](https://gitlab.haskell.org/haskell/ghcup-hs/-/tree/master/www)
 
-### Cabal instructions
-Just run `cabal v2-build` to build or `cabal v2-run` to run, and `cabal v2-run -- build` to actually build the site.
+### Nix
 
-### Nix instructions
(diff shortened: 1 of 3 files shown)


tomjaguarpaw:
Is it OK if we don't de-emphasise the cabal instructions so much? I (and I suspect many other contributors) would prefer not to have to get started with Nix just to edit this site.

tomjaguarpaw:
Regarding factoring out `buildAndWatch` that looks like a good idea.

tomjaguarpaw (approved):

rebeccaskinner:
> Is it OK if we don't de-emphasise the cabal instructions so much? I (and I suspect many other contributors) would prefer not to have to get started with Nix just to edit this site.

I've updated the `buildAndWatch` script to be a bit smarter and to try to use nix if it's available, and otherwise to fallback to using cabal. I updated the documentation to put nix and cabal on equal footing, but to prioritize using the script, since I think that's going to be the easiest route to onboarding contributors.

tomjaguarpaw:
Looks great, thanks!

PR #95, opened 2021-06-18
Haskell company quotes

rebeccaskinner:
Draft:
This is currently a draft PR. It still needs frontend work. It will allow us to add testimonials to the site. Do not merge, as right now it has some ugly visual styling for demo purposes.


builder/site.hs
```diff
@@ -1,15 +1,39 @@
...
+    compile copyFileCompiler
+
+  match "testimonials/*.yaml" $ do
+    compile parseTestimonialCompiler
+
+  create ["testimonials.json"] $ do
+    route idRoute
+    compile $ do
+      testimonials <- loadAll @Testimonial "testimonials/*.yaml"
+      item <- makeItem $ (BL.unpack . encode . map itemBody) testimonials
```
tomjaguarpaw, line 33: I like the use of `(.)`.  Why not go all the way and use `(makeItem . BL.unpack . encode . map itemBody) testimonials`?

tomjaguarpaw:
Looks very nice!

builder/site.hs
rebeccaskinner: ah good call, no idea why I didn't do that in the first place TBH

testimonials/README.md
```diff
@@ -0,0 +1,65 @@
+# Haskell.org Testimonials
+
+Is your company using haskell? Do you want to tell the world about it? This is
```
ulysses4ever, line 3: Should it be Haskell rather than haskell here and couple other places?

rebeccaskinner: Yeah, good call, it probably should. I'll update it. Thanks!

tomjaguarpaw:
This is great, but a bit compressed at standard zoom level on my browser (Firefox).

screenshot

rebeccaskinner:
2021-07-11-192514_1800x2113_scrot

I've done a bit to fix some of the styles so that there shouldn't be visible scrollbars on firefox at almost all viewport sizes (there's a small range of sizes between typical mobile and typical desktop sizes where there are scrollbars unfortunately, but I think this catches almost all normal browser viewport sizes). It should work with or without js enabled and look decent now.

I've tested on Linux, mostly in firefox, and a bit in Chrome. If anyone has access to a mac and can test with safari, or has an environment setup to easily test on simulated phone browsers that would be very helpful.

tomjaguarpaw:
Now it looks like this. It's weird, because as I make my screen narrower more entries appear (five total)! But my screen is particularly narrow, so this probably isn't such a big deal. I'd certainly suggest deploy first and fix later, unless you're particularly keen to fix first.

screenshot

tomjaguarpaw (approved):

rebeccaskinner:
@tomjaguarpaw hmm, I have a very high resolution monitor- let's get a couple of other folks to review as well and see whether the predominate experience is what you are seeing or what I'm seeing. I don't mind deploying if we'd only see in some odd scenarios, but I'd rather not deploy if this is going to be the common experience.

tomjaguarpaw:
Same problem on Chromium (my previous screenshot was Firefox). Both on Linux.

With a narrow page width there is one testimonial (which is fine)

screenshot

a bit wider and there are five

screenshot

a bit wider still there are *six*

screenshot

yet wider and there are four (which is fine)

screenshot

So something funny is going on.

rebeccaskinner:
I'll mess with it a bit more tonight and see, my guess is that this has something to do with the responsive CSS stuff, which is so far out of my area of expertise it's not even funny but I might be able to either flail about a bit and get it fixed, or else pester some of my frontend-friends into helping me out.

TikhonJelvis:
Minor point: on my phone, the left arrow is cut off completely, and the right arrow is probably a bit too close to the right side of my screen.

Here's a screenshot:

image

TikhonJelvis:
Also, would be cool if I could swipe through quotes with my finger—I guess that's something I've seen on other mobile sites, because I tried it without even thinking about it and was a bit surprised it didn't work.

rebeccaskinner:
Huh, I could have sworn that at one point swiping like that on mobile did work- I wonder if I broke it when I added the code to shuffle the order of the quotes.

tomjaguarpaw:
This latest version works great on my browser!

rebeccaskinner:
@TikhonJelvis I believe that this last set of changes should fix the mobile issues. I have verified on iOS that it's working correctly now.

testimonials/mercury.yaml
```diff
@@ -0,0 +1,6 @@
+---
+companyName: Mercury
+logoURL: logos/mercury.png
+shortTestimonial: |
+  Mercury is a bank for businesses. We use Haskell to meet our
+  customers high standards for correctness and security,
```
tomjaguarpaw, line 6: Should be `.` instead of `,`?

tomjaguarpaw:
> This latest version works great on my browser!

Still works great

builder/site.hs
```diff
@@ -1,15 +1,39 @@
...
-import           Data.Monoid           ((<>))
+import           Control.Monad
+import           Data.Aeson
+import qualified Data.ByteString.Lazy          as BL
+import           Data.Monoid                   ((<>))
 import           Data.Time.Calendar
 import           Data.Time.Clock
 import           Hakyll
+import           Hakyll.Core.Compiler
+import           Hakyll.Core.Compiler.Internal
```
tomjaguarpaw, line 12: I'm feeling uneasy about importing internal modules. Is there really no way to get these from a supported interface?

builder/site.hs
```diff
@@ -38,29 +67,34 @@ main = mkContext >>= \ctx -> hakyll $ do
     compile templateCompiler
 
 
+parseTestimonialCompiler :: Compiler (Item Testimonial)
+parseTestimonialCompiler = do
+  identifier  <- getUnderlying
+  provider    <- compilerProvider <$> compilerAsk
```
tomjaguarpaw, line 73: Internal import is used here. Can these be found in a non-internal module?

rebeccaskinner: If we update to the latest hakyll this is part of the public API. I'm happy to do it, but I wonder if maybe we'd be better off merging this as-is and then taking on the update as a second piece of work afterwards? I'd be happy to take on the update as well.

tomjaguarpaw: I'll leave the decision up to you. Personally I'm happy to merge now and update later.

testimonials/mercury.yaml
rebeccaskinner: yep, thanks for the catch. fixed.

builder/site.hs
```diff
@@ -1,15 +1,37 @@
...
+    compile copyFileCompiler
+
+  match "testimonials/*.yaml" $ do
+    compile parseTestimonialCompiler
+
+  create ["testimonials.json"] $ do
+    route idRoute
+    compile $ do
+      testimonials <- loadAll @Testimonial "testimonials/*.yaml"
+      item <- (makeItem . BL.unpack . encode . map itemBody) testimonials
```
TikhonJelvis, line 31: I'm not following this code exactly: why do we need the `BL.unpack`? What are we using a `[Word8]` for?

css/hl.css
```diff
@@ -3374,7 +3375,15 @@ a.badge:hover, a.badge:focus {
     margin-right: auto; }
   .thumbnail .caption {
     padding: 9px;
-    color: #333333; }
+    color: #333333;
```
TikhonJelvis, line 3378: This is more a note for the future than for this PR, but we could definitely start using CSS variables for colors like this.

css/hl.css
```diff
@@ -4608,6 +4617,47 @@ pre .hs-keyglyph {
     -webkit-flex: 0 1 auto;
         -ms-flex: 0 1 auto;
             flex: 0 1 auto; }
-    .row-flex > div[class*='col-'] .thumbnail {
+  .row-flex > div[class*='col-'] .thumbnail {
       width: 100%; }
 
+  .glider,.glider-contain{margin:0 auto;position:relative;}
```
TikhonJelvis, line 4623: Would it be worth getting a CSS formatter to make the CSS a bit more uniform and readable? I haven't used one before, so I can't recommend any particular tool.

TikhonJelvis (approved):
Looks great overall and I love the README file. I left a couple of questions/comments, but nothing that would get in the way of merging this PR as-is.

builder/site.hs
rebeccaskinner: We load all of the testimonials, then serialize them to json and convert the json to raw bytes that we can save to disk. There are a bunch of other ways we might have done this. It could probably use a comment. Since we're intending to upgrade hakyll and refactor some of this code, I'll ensure we more thoroughly comment whatever the resulting refactored code is.

css/hl.css
rebeccaskinner: Yeah, I agree we probably should look into that if we can find a good one.

PR #97, opened 2021-06-28
documentation: remove Hayoo

ulysses4ever:
It had been down for a couple years now


gbaz:
I also did this in #99 btw, didn't realize there was overlap

tomjaguarpaw (approved):
Thanks for spotting this! Will probably just do this via https://github.com/haskell-infra/www.haskell.org/pull/99 though.

(Again I'm curious why linkchecker didn't spot it, cc @TikhonJelvis. Perhaps it doesn't work if a site times out?)

ulysses4ever:
Thanks all! I'm cool with just getting #99 in.

PR #101, opened 2021-07-05
Add testimonial section to index

AlbertVC:
Hello I am a developer of strivelabs, we have developed the HTML and style of a carousel for the testimonal section on the index. The attachments should let you see the result

https://user-images.githubusercontent.com/26113462/124448688-01779300-dd83-11eb-83b9-819ce8efd300.mp4

Testimonial desktop
testimonial mobile

### DIFF
--- a/css/hl.css
+++ b/css/hl.css
@@ -3374,7 +3374,15 @@ a.badge:hover, a.badge:focus {
     margin-right: auto; }
   .thumbnail .caption {
     padding: 9px;
-    color: #333333; }
+    color: #333333; 
+  }
+  .thumbnail .testimonial-caption {
+    padding: 9px;
+    color: #333333; 
+    height: 250px;
+    overflow-y:auto;
+    word-wrap: break-word;
+  }
 
 a.thumbnail:hover,
 a.thumbnail:focus,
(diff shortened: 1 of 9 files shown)


rebeccaskinner (commented):
This is looking great! I'm going to have a proper look at it tomorrow but it's looking great at first glance- thanks so much for the help!

rebeccaskinner (changes requested):
I think this is really good, but is there any way we could update the stylesheet so that if a user has javascript disabled the testimonials still fill the width of the content area, and make sure that the height of all of the text boxes is uniform.

Thanks again so much for the help with this!
2021-07-07-182615_1916x2113_scrot

2021-07-07-182836_1916x1054_scrot

AlbertVC:
Hello! Thanks for the feedback, regarding the same height of the text bosses, do you prefer to hide the overflow or to add a scroll to see it? The attachments should let you see the look of both options.
We are glad to help!
scroll
hidden

AlbertVC:
> I think this is really good, but is there any way we could update the stylesheet so that if a user has javascript disabled the testimonials still fill the width of the content area, and make sure that the height of all of the text boxes is uniform.
> 
> Thanks again so much for the help with this!
> 2021-07-07-182615_1916x2113_scrot
> 
> 2021-07-07-182836_1916x1054_scrot

Hello! We are glad to be able to help with this project, we have changed the style so now text boxes should have the same height also if the text inside a box is too long a horizontal scroll bar will appear. Please take a look to the attachment to see the results
new style

rebeccaskinner (approved):
This is looking great to me

PR #103, opened 2021-07-10
List subsites

tomjaguarpaw:
It would be very helpful to collect the list of `www.haskell.org` subsites in one place. Eventually I would like to gather more information about their source and owners.

The table is split into two sections: subsites that are reasonably up-to-date and subsites that aren't. Within the sections the subsites appear in alphabetical order..

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,6 @@ This is the website for www.haskell.org built as a hakyll static site, which bui
 
 * [The PR policy of this repository](https://github.com/haskell-org/committee/blob/main/proposals/0003-pr-process.md)
 
-### Subsites
-
-Not all subsites of www.haskell.org are built from this repository.
-Some of the others are
-
-* www.haskell.org/cabal (built from [cabal-website](https://github.com/haskell/cabal-website))
-* www.haskell.org/platform (built from [haskell-platform](https://github.com/haskell/haskell-platform/tree/master/website))
-* www.haskell.org/ghcup (build from [ghcup-hs](https://gitlab.haskell.org/haskell/ghcup-hs/-/tree/master/www)
-
 ### Contributing Changes
 
 The easiest way to make changes is to use the `buildAndWatch` script and then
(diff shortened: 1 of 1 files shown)


gbaz:
The haddock, happy and alex sites should be owned by the respective maintainers for the three projects.

The definition subsite should be owned by the same people as the onlinereport subsite -- which would be the haskell prime committee, except as we know that's perpetually stalled...

ghc-perf was the purview of joachim breitner but can probably go.

The following pages are all basically archival, but here is who I would confer with on them:

hugs and nhc98 are archival pages, mainly frozen in time. I suppose mark jones and malcolm wallace would be the points of contact.

The haskell workshop page is also an archival page but ultimately the responsibility of the haskell symposium organizers.

The tutorial page is the legacy of Paul Hudak, and I suppose one would want to confer with his former students and collaborators about it if necessary.

ulysses4ever:
I'm concerned quite a bit about the above mentioned "archival" status of some of the subsites. It greatly confuses a lot of newcomers that they get references to discountinued tools and tutorials right under haskell.org without any warning about that. Keep in mind that these "subsites" may be not very discoverable from the root of haskell.org, but still are quite hight on google search results. 

I'm wondering if haskell.org could move those archival subsites from say `/tutorial` to `/archive/tutorial`. That would make our more obvious the status. 

Better yet would be to contact the webmasters of the discountinued subsites and ask them to put a warning on their sites. But that's may be harder to implement.

gbaz:
A little investigation reveals this concern to be largely already addressed.

Hugs does have an archival notice I worked out with approval of mark jones back in the day. The nhc98 page clearly signposts it was last released 10 years ago, which should be a good tipoff for users. Also I hardly suspect stray googlers will go specifically looking for nhc98.

The tutorial is otherwise known as the "gentle introduction" and its a classic work that is still (correctly) linked to from some places. It also clearly signposts when it was last published, so users will not be confused.

Which leaves the symposium site, which also has a clear notice that it is superseded by the haskell workshop and has a link to the more current information.

In my experience helping users with many questions over the years, I have seen many things confuse them. Among those things, I have _not_ seen somebody get confused by encountering e.g. the nhc98 page.

I think a better focus would be on helping the haddock team maintain its website, which I _have_ seen people get confused by outdated information from, on the other hand.

ulysses4ever:
Fair enough. Imo the "last modified" date is not at all enough to make it clear that, say, the tutorial has mostly historical significance. If something has the haskell.org/tutorial url and was last modified in 2000 it may mean that the language has not been updated much since then and the tutorial is still relevant. In general, depending on a language this may or may not be a valid guess. But you can't know without having some context, which newcomers most often lack.

ketzacoatl:
This seems like a useful change, is there anything blocking it, or is it more accurate to close as "won't fix"?

tomjaguarpaw:
> This seems like a useful change, is there anything blocking it, or is it more accurate to close as "won't fix"?

Nothing is blocking it except my lack of time. It's on my stack of things to do, not very high up.

tomjaguarpaw:
> ghc-perf was the purview of joachim breitner but can probably go.

@gbaz: Does this mean that subsite can be removed?

I have incorporated all the changes suggested by @gbaz (I hope).

Waiting for: review (cc @TikhonJelvis @rebeccaskinner, @jaspervdj, other committee members)

rebeccaskinner (approved):
This is a big improvement. Sorry for the delayed approval.

PR #104, opened 2021-07-14
resize testimonials to 300px wide

imalsogreg:
Before this change, testimonials would be squeezed uncomfortably tight on narrow viewports. This PR sets the width to 300px unconditionally, and also deletes a couple of confusing static files related to the `glider.js` widget but apparently not used.

After the resize:

Screen Shot 2021-07-13 at 10 51 04 PM

Screen Shot 2021-07-13 at 10 51 18 PM

cc @rebeccaskinner

### DIFF
--- a/js/glider.js
+++ b/js/glider.js
@@ -1,6 +1,0 @@
-window.addEventListener('load', function(){
-    new Glider(document.querySelector('.glider'), {
-      setting: setting-value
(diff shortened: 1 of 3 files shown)


rebeccaskinner (approved):
This looks great to me, thanks for the help!

PR #105, opened 2021-07-15
Removed meetup group that no longer exists

ackinc:


tomjaguarpaw (approved):
Thanks!

TikhonJelvis (approved):
Thanks for catching that!

PR #106, opened 2021-07-15
documentation: add reference to https://ghc.dev/

ketzacoatl:


tomjaguarpaw:
Since this is for developers of GHC rather than users it seems that a more appropriate place would be the [GHC developers wiki](https://gitlab.haskell.org/ghc/ghc/-/wikis/home).

TikhonJelvis:
Do we want a link to the GHC developer's wiki itself? Pointing people to resources for GHC development could be good, but it feels like we already have so many links on this documentation page that it's pretty overwhelming/hard to navigate.

ketzacoatl:
I don't know how anyone would find the GHC developers wiki, or this ghc.dev, without it being part of a central hub/authority on "haskell documentation". I consider haskell.org to be the place for that, else I don't know why we have the site, or what purpose it serves (except to confuse us into thinking we will get the info we need to get started with X).

Just my two cents, feel free to pass it by.

TikhonJelvis:
Right, if the resource page were organized better—or even split into multiple pages—it could do a better job presenting a large number of resources. What should we do about that?

One self-contained step would be putting together a high-level "how to contribute to GHC" page with some encouragement and links to key resources. A page like that would be the natural place for links to the GHC wiki, ghc.dev... etc. I imagine the GHC team already has a lot of resources like this, so it's a matter of introducing that and linking from the main haskell.org page. Eventually this would also be a model for further organizing the resources we present on the website.

Looking at ghc.dev, seems like there would be overlap between the two, but ghc.dev looks like it's oriented to people who have a basic idea of what's going on and need pointers to specific resources. Is there an overview for GHC development for somebody who is just thinking about getting started?

tomjaguarpaw:
I'm in favour of a *single* link (perhaps displayed prominently under a large title) to an authoritative source of information relating to developing GHC. That source should be the responsibility of the GHC team. If the GHC team wants that source to be a page under haskell.org then that's probably fine, but I suspect it would be better to keep it under, and I suspect they would *want* to keep it under, their GitLab account. Whether they want to link to ghc.dev from there is up to them.

Why am I discouraging having a link to ghc.dev here? Because not only should we avoid providing too little information, we should also avoid providing too *much* information. I can't imagine that any user would be better served by a link to ghc.dev than a prominent link to a separately-maintained collection of resources related to developing GHC. If anyone has a contrary point of view then I would be interested to hear more.

gbaz:
If I google "contribute to ghc" the first link is the following, which is an officially maintained ghc contribution guide

https://gitlab.haskell.org/ghc/ghc/-/wikis/contributing

That page is linked from the central ghc wiki, which has other ghc information as well:

https://gitlab.haskell.org/ghc/ghc/-/wikis/home

A single link to the ghc wiki as a whole (which has information for users and developers both) could make sense, in my opinion.

rebeccaskinner (approved):
This looks like a good change to me- thanks!

PR #113, opened 2021-07-18
Changed testimonials/mercury.yaml

uzair-javed-1:
in \testimonials\mercury.yaml file , Customer is alreay a Plural/Noun so that we can't use it customers or customer's . it's a Grammer mistake , so that i replaced customer's with customer.


uzair-javed-1:
in \testimonials\mercury.yaml file , Customer is alreay a Plural/Noun so that we can't use it customers or customer's . it's a Grammer mistake , so that i replaced customer's with customer.

PR #115, opened 2021-07-19
Haskell company quotes

romefeller:
Inroducing IOHK to the testimonials page.


tomjaguarpaw (approved):

tomjaguarpaw:
We might want to ask them to flesh this out more. It's a bit vague.

> when it comes to programming languages work there is no competitor to Haskell.

TikhonJelvis:
I believe a lot of their PL-oriented work is either open source or at least public, right? A quote referencing a specific project or effort would be more effective—"show" rather than "tell".

rebeccaskinner (approved):
If IOHK is happy with this as their quote, I think it's good. Many of the submissions we got were too long to fit comfortably on the screen, I think this is about the length that I originally had in mind when we started this.

romefeller:
@TikhonJelvis can you merge it?

TikhonJelvis:
Sure, let's go for it. More specific/compelling quotes would be nice, but sounds like it isn't worth the hassle right now. We can revisit this fish the line when it's time to update/refresh the testimonials.

PR #116, opened 2021-07-22
Adds CSS Linter and Formatter

### LINKED ISSUE #110: Add a CSS Auto-Formatter
We should add a CSS auto-formatter to the nix environment, and format the existing CSS so that it looks nice. Consider possibly using a commit hook to auto-format new commits.

mechamobau:
In this PR is added a CSS Linter and Formatter to the project and created a step to verify the CSS on new Pull Requests.

I've added the Node.js environment to the action because I have no familiarity with Nix environments, so I'm open to change that according to any given suggestions.

To run that locally it's needed to have Node.js installed and run the following command inside the root directory:

```bash
$ npx stylelint 'css/*.css' --fix
```

closes #110

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,11 @@ jobs:
           name: haskell-org
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
+      - name: Lint CSS
+        uses: actions/setup-node@v2
+        working-directory: './'
+        run: npx stylelint 'css/*.css' --config .stylelintrc.json --fix
+
       - name: Build Site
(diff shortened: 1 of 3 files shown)


TikhonJelvis:
Thanks for the config. It would be great to do this through Nix rather than installing NPM separately.

I might give it a try myself, but don't have the time/attention at the moment. In the meantime, please feel free to give it a go yourself if you're interested in learning!

TikhonJelvis:
Looks like this will need some Nix work to be useful, and I probably won't get around to it any time soon. Going to close the PR for now, and we can always pick it up again later if we want linting/minimization for our CSS in the future.

PR #117, opened 2021-07-23
Update Foxhound Systems testimonial

charukiewicz:
I'm one of the Co-Founders at Foxhound Systems. I didn't really provide a great quote in my original email to the Haskell committee, since I thought only company name and logo would be included. I provided a better quote in this PR, one that emphasizes why Haskell specifically is a great language.


tomjaguarpaw (approved):
Thanks!

TikhonJelvis (approved):

PR #119, opened 2021-07-28
Update Fun. Design and Arch. book link

noughtmare:
The first edition is no longer being sold, so this new link points to the second edition.


tomjaguarpaw (approved):
Thank you!

TikhonJelvis (approved):

PR #123, opened 2021-07-29
Add company backlinks to testimonial logos

### LINKED ISSUE #118: Could testimonials link back to the company websites?
So cool to see the testimonials on the live site! Any chance they could link back to the company home pages? I'd be happy to send a pull request myself if so.

aviaviavi:
Closes #118 

This change makes the logos in the testimonial section link back to the companies that gave them. I've added links to all of the companies, so in addition to the merits of the code change, it would be probably be a good idea for someone to double-check I got all of the links correct.

### DIFF
--- a/builder/Testimonial.hs
+++ b/builder/Testimonial.hs
@@ -13,6 +13,7 @@ data Testimonial = Testimonial
   { companyName      :: String
   , logoURL          :: String
   , shortTestimonial :: String
+  , companyURL       :: String
   } deriving (Eq, Show, Generic)
 
 instance Binary Testimonial where
@@ -20,8 +21,9 @@ instance Binary Testimonial where
     put companyName
     put logoURL
     put shortTestimonial
+    put companyURL
 
-  get = Testimonial <$> get <*> get <*> get
+  get = Testimonial <$> get <*> get <*> get <*> get
 
 instance Writable Testimonial where
   write path Item{..} =
--- a/builder/site.hs
+++ b/builder/site.hs
@@ -109,4 +109,5 @@ testimonialContext =
   mconcat [ field "companyName" (pure . companyName . itemBody)
           , field "logoURL" (pure . logoURL . itemBody)
           , field "shortTestimonial" (pure . shortTestimonial . itemBody)
+          , field "companyURL" (pure . companyURL . itemBody)
           ]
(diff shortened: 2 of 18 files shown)


tomjaguarpaw (approved):

tomjaguarpaw:
Thanks! I like it and I'm happy for it to be merged as it is, but I have two suggestions:

1. Don't use `target=_blank`. We're not trying to optimise some user-engagement metric and keep visitors on our page. Let's just use a normal link and let visitors choose to open the links in a new tab or not depending on their own personal preference. Personally I *really* hate it when websites change default link behaviour like this.

2. Put the link on the company name as well as the image.

I'm curious what @TikhonJelvis and @rebeccaskinner think.

aviaviavi:
Fair enough :) always good to hear people's preferences on usability stuff like that 

I think these changes sound good, will do 👍

aviaviavi:
Updated with the requested changes!

rebeccaskinner (approved):
I think this looks great!

rebeccaskinner:
> Thanks! I like it and I'm happy for it to be merged as it is, but I have two suggestions:
> 
>     1. Don't use `target=_blank`.
>        We're not trying to optimise some user-engagement metric and keep visitors on our page. Let's just use a normal link and let visitors choose to open the links in a new tab or not depending on their own personal preference. Personally I _really_ hate it when websites change default link behaviour like this.
> 
>     2. Put the link on the company name as well as the image.
> 
> 
> I'm curious what @TikhonJelvis and @rebeccaskinner think.

I don't really have strong feelings one way or another, but I have a slight preference toward opening the links in a new tab. It fits better with my mental model of someone being curious about haskell, seeing the testimonials, and wanting to open up a few of them in new tabs to learn more about the kind of companies that are using haskell.

That said, I certainly don't think it's worth blocking the PR one way or another, and it's easy enough to change either way if people don't like what we pick.

TikhonJelvis:
When I browse myself, I strongly prefer normal links. However, I have no idea whether this matches other people's expectations!

As a point of comparison, I looked at [Rust's site](https://rust-lang.com) which also has testimonials with links as logos, and they open in the same page. That said, they also point to specific talks/etc rather than company pages, so it's not a 1:1 comparison.

TikhonJelvis (approved):
I don't have time at the moment to build/preview this PR locally myself, but looks great based on the code changes, and I'm in favor of merging since @tomjaguarpaw and @rebeccaskinner looked it over :).

PR #93, opened 2021-06-12
Simplify download page further

hasufell:
Since ghcup now runs on pretty much any platform and supports stack, I leave this here for discussion of further download page simplification.

----

It should be noted that there is a minor caveat of ghcup's stack handling: of course it doesn't interact well with `stack upgrade`, which overwrites binaries (no matter the name of the invoking executable). Users will be notified to use ghcup mechanism for such updates.

PR #125, opened 2021-08-11
Update downloads page

Bodigrim:
Message after @tomjaguarpaw's changes

The specifics of this change are to replace Chocolatey with ghcup as the preferred installation method of GHC, cabal-install and haskell-language-server on Windows. The overall effect is that the install instructions become common over Windows, Linux and MacOS.

In my view this makes something better without making anything worse.  However, the Haskell.org committee can't proceed to merge this without broad community input.

Counterpoints:

* @parsonsmatt suggests that ghcup on Windows is not sufficiently well-tested.

----

@bodigrim's original message

Since #93 has descended into chaos, I’ll take a risk to suggest an alternative, more conservative approach, which merely mentions that ghcup supports Windows and can install Stack, but does not go any further. I hope it could be more palatable, but if I’m wrong feel free to close, no hard feelings on my part.

This PR still offers a considerable simplification and unification, because installation instructions for all major platforms are now uniform.

### DIFF
--- a/downloads.markdown
+++ b/downloads.markdown
@@ -16,35 +16,16 @@ This page describes the installation of the Haskell toolchain, which consists of
 
 *   [haskell-language-server](https://github.com/haskell/haskell-language-server) (optional): A language server for developers to integrate with their editor/IDE
 
-Select your platform to get more specific installation instructions:
-
-*   [Linux, OS X and FreeBSD](#linux-mac-freebsd)
-
-*   [Windows](#windows)
-
-Alternative methods to install GHC are listed [here](#ghc-install-manual).
-
-* * *
-
-## Platform specific instructions
-
-### Linux, OS X and FreeBSD { #linux-mac-freebsd }
+## Installation instructions
 
 1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
 2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
-
-### Windows { #windows }
-
-1. Install GHC and cabal-install via Chocolatey
-	- [Configure Chocolatey](https://chocolatey.org/install) on your machine
-	- At an elevated command prompt, run `choco install haskell-dev`, followed by `refreshenv`.
-2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#windows)
 
 * * *
 
 ## Alternative installation options { #ghc-install-manual }
 
-*   [Using a package manager on linux](#package-manager)
+*   [Using a package manager](#package-manager)
 
 *   [Official bindists](#bindists)
 
@@ -52,7 +33,9 @@ Alternative methods to install GHC are listed [here](#ghc-install-manual).
 
 *   [Other options](#other-options)
 
-### Using a package manager on Linux { #package-manager }
+### Using a package manager { #package-manager }
+
+#### Linux
 
 Refer to your distribution package manager documentation. For convenience, below are a few distribution specific instructions, outlining 3rd party repository use as well.
 
@@ -162,6 +145,11 @@ sudo emerge --ask dev-lang/ghc dev-haskell/cabal-install
 ```
 
 </div>
+
+#### Windows
+
+1. [Configure Chocolatey](https://chocolatey.org/install) on your machine
+2. At an elevated command prompt, run `choco install haskell-dev haskell-stack`, followed by `refreshenv`.
 
 ### Official bindists { #bindists }
 


downloads.markdown
```diff
@@ -16,45 +16,28 @@ This page describes the installation of the Haskell toolchain, which consists of
...
-
-*Note: Apple Silicon is not yet supported. You will have to install GHC in `arch -x86_64` compatibility mode using rosetta. See [here](https://derflounder.wordpress.com/2020/11/17/installing-rosetta-2-on-apple-silicon-macs/) and [here](https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/101) for more pointers.*
-
-### Windows { #windows }
-
-1. Install GHC and cabal-install via Chocolatey
-	- [Configure Chocolatey](https://chocolatey.org/install) on your machine
-	- At an elevated command prompt, run `choco install haskell-dev`, followed by `refreshenv`.
-2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#windows)
+1. Use [GHCup](https://www.haskell.org/ghcup/) to install
```
hasufell, line 21: This suggests to install cabal/GHC via ghcup on windows and lists chocolatey as an alternative. Is that intentional?

hasufell: > Chocolatey was moved below under platform-specific instructions for package managers. If this move is controversial, I'm happy to restore it back, this was done in the interest of clarity and brevity, not because of partisan preferences.

True, I agree here. But the argument from @parsonsmatt was that neither windows nor stack support are sufficiently "battle-tested".

downloads.markdown
```diff
@@ -16,45 +16,28 @@ This page describes the installation of the Haskell toolchain, which consists of
...
-* * *
-
-## Platform specific instructions
-
-### Linux, OS X and FreeBSD { #linux-mac-freebsd }
-
-1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
-2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
-
-*Note: Apple Silicon is not yet supported. You will have to install GHC in `arch -x86_64` compatibility mode using rosetta. See [here](https://derflounder.wordpress.com/2020/11/17/installing-rosetta-2-on-apple-silicon-macs/) and [here](https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/101) for more pointers.*
```
tomjaguarpaw, line 36: Do we really want to remove this warning?

hasufell: Yes, because it is perfectly supported right now!

tomjaguarpaw (approved):
The specifics of this change are to replace Chocolatey with ghcup as the preferred installation method of GHC, cabal-install and haskell-language-server on Windows. The overall effect is that the install instructions become common over Windows, Linux and MacOS. Great!

Chocolatey is moved to "Using a package manager". We should check how @Mistuke feels about this. Still, I am strongly in support of moving towards a common installation method.

I approve this PR. I think it makes some things better without making anything worse. In particular it doesn't impinge on Stack at all. I think a few others from the committee should have a look.

jneira:
The change looks good and it will help to increase ghcup adoption for stack and windows users so :+1:

downloads.markdown
parsonsmatt: Yeah this PR has the same problem as the other one, as far as I am concerned

I think the safest option is to add `ghcup` as an option under Windows, but not make it the preferred/recommended/"most obvious choice" until it's been in the wild for a bit more time.

Bodigrim: @parsonsmatt I'm sympathetic to your view in a context of the previous PR. However, here no installation method is removed, only one added. 

Note that ghcup is actually the **only** tool, fully capable of installing Haskell toolchain (as defined above). So unless there is a specific known issue with regards to Windows support in ghcup, I think it's worth to list it as proposed. But - again - I'm happy to leave it for haskell.org committee to deliberate and decide.

parsonsmatt: This PR does quite a bit more than add an option to the list, which would be completely fine IMO.

The first presented option that is *relevant* to a user should be a well-tested "blessed" choice, with extremely little chance of failure or problems. The current page structure accomplishes this. The user selects their platform, and is given the 'blessed' path - `ghcup` for Linux/OSX/FreeBSD and `chocolatey`+`stack` for Windows.

The simplest change that adds `ghcup` as an option is to just add it as an option in the current Windows section.

```diff
 ### Windows { #windows }
 
 1. Install GHC and cabal-install via Chocolatey 
 	- [Configure Chocolatey](https://chocolatey.org/install) on your machine
 	- At an elevated command prompt, run `choco install haskell-dev`, followed by `refreshenv`.
 2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#windows)
+3. `ghcup` recently acquired Windows support. To install `ghcup`, follow the instructions [here](wherever this needs to be)
```

This change accomplishes pretty much what everyone wants: `ghcup` gets mentioned under Windows support, `ghcup` gets exposure and more users and more testers, the `stack` team's concerns about `ghcup` as an installation method for `stack` are satisfied, my concerns about new-users being funneled into relatively less-tested workflows is abated. 

In ~6 months (or whatever timeline works for folks), when there is more confidence in `ghcup`'s Windows support, then we can switch to this unified installer plan. Then we have one main "How to get Haskell" section (heck yes) and alternatives for different platforms listed below.

hasufell: > This PR does quite a bit more than add an option to the list, which would be completely fine IMO.
> 
> The first presented option that is *relevant* to a user should be a well-tested "blessed" choice, with extremely little chance of failure or problems. The current page structure accomplishes this. The user selects their platform, and is given the 'blessed' path - `ghcup` for Linux/OSX/FreeBSD and `chocolatey`+`stack` for Windows.
> 
> The simplest change that adds `ghcup` as an option is to just add it as an option in the current Windows section.
> 
> ```diff
>  ### Windows { #windows }
>  
>  1. Install GHC and cabal-install via Chocolatey 
>  	- [Configure Chocolatey](https://chocolatey.org/install) on your machine
>  	- At an elevated command prompt, run `choco install haskell-dev`, followed by `refreshenv`.
>  2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#windows)
> +3. `ghcup` recently acquired Windows support. To install `ghcup`, follow the instructions [here](wherever this needs to be)
> ```
> 
> This change accomplishes pretty much what everyone wants: `ghcup` gets mentioned under Windows support, `ghcup` gets exposure and more users and more testers, the `stack` team's concerns about `ghcup` as an installation method for `stack` are satisfied, my concerns about new-users being funneled into relatively less-tested workflows is abated. 
> 

This is specifically what I do not want: add more options/choice for the user.

The reason the DL page was confusing is exactly this: listing all sorts of installation methods, leaving the user confused.

Ultimately, it should be one single instruction. If that's not feasible, it should be one single instruction per platform. Alternatives are to be listed as a footnote or not at all.

hasufell:
I'm in favor of this change, because windows support is something I have full control over and can adjust and fix freely.

Fixing integration issues with stack may require collaboration, which is in limbo right now.

tomjaguarpaw:
> windows support is something I have full control over and can adjust and fix freely.

This is an important point. I would augment it with the observation that we are grateful that @Mistuke developed and supports the Chocolately installer despite being busy with many other things. Regarding ghcup Windows support, based on @hasufell's track record, we can be confident he will be extremely responsive and effective in supporting it, should any problems arise. (Because it is written in Haskell it is easier for *others* to support too, should the need arise.)

tomjaguarpaw:
@Bodigrim I rebased your branch on `master` and made some cosmetic changes: https://github.com/haskell-infra/www.haskell.org/pull/129. Could you please take a look and if you approve force push it on top of this branch?

hasufell:
Wrt chocolatey, I created https://gitlab.haskell.org/haskell/ghcup-hs/-/issues/208

Bodigrim:
@tomjaguarpaw sorry, I'm AFK until next week. Feel free to push force the rebased version over my branch, you should have necessary permissions.

tomjaguarpaw:
I have force pushed my cosmetic changes and Julian's suggestions from https://github.com/haskell-infra/www.haskell.org/pull/129 to this branch.

hasufell (approved):

tomjaguarpaw:
I've simplified this further, minimizing the diff (in part by splitting out https://github.com/haskell-infra/www.haskell.org/pull/130, which is logically separate).

Bodigrim:
@tomjaguarpaw any chance to merge this soon?

tomjaguarpaw:
Since this counts as a "controversial change" it will need a majority of the Haskell.org committee to approve it before merge. The committee is required to take into account the interests of the full range of the Haskell community. The effect of this PR is to promote ghcup and de-emphasise Chocolatey so the opinion of Chocolately users and developers is particularly welcome. Other stakeholders should have their input too.

Another concern, raised by @parsonsmatt, was whether ghcup on Windows is sufficiently well-tested. ghcup has had Windows support for a while now. Perhaps we have enough evidence to say yes it is sufficiently well-tested.

hasufell:
> The effect of this PR is to promote ghcup and de-emphasise Chocolatey so the opinion of Chocolately users and developers is particularly welcome.

Note that ghcup prints a message on windows if it detects chocolatey and allows the user to stop and use that instead.

> whether ghcup on Windows is sufficiently well-tested. ghcup has had Windows support for a while now. Perhaps we have enough evidence to say yes it is sufficiently well-tested.

@jneira has used it extensively there, any opinion?

Bodigrim:
@tomjaguarpaw all right, what would it take to trigger Haskell.org committee vote on this? The PR is open for 7 weeks, I would like to have it resolved one way or another.

jneira:
> @jneira has used it extensively there, any opinion?

I have to say that i cant imagine manage the haskell toolchain without ghcup now. I was doing it manually before that and was cumbersome. Unify the haskell toolchain setup process for all main platforms is a must imo and ghcup is the working alternative 
right now.

More importantly, the issues reproduced in windows has been resolved diligently by ghcup developers and the project is running its test suite in windows to avoid regressions. And it is previsible it will continue that way.

tomjaguarpaw:
The next meeting of the committee is in two weeks. It is also possible to hold the vote asynchronously. I will communicate my support for this PR to the committee.

alirezaghey:
I'm a nobody and have little to now experience but haven't seen but good from `ghcup`, although I haven't used it on Windows. It made the setup very straightforward for me and I wished I had found it earlier.

cartazio:
i support @hasufell's tooling as the recommended general user installation for all platforms he feel confident about supporting.

tomjaguarpaw:
I sought community input on [Discourse](https://discourse.haskell.org/t/seeking-community-input-on-change-to-haskell-org-downloads-page/3381), [Reddit](https://www.reddit.com/r/haskell/comments/pzvo9w/seeking_community_input_on_change_to_haskellorg/) and [haskell-cafe](https://mail.haskell.org/pipermail/haskell-cafe/2021-October/134661.html). There was significant support for this change. No one objected to Chocolatey being de-emphasised. The Slack maintainership have told me they have no objection to this change.

I strongly recommend this PR to the committee.

tomjaguarpaw:
cc @TikhonJelvis @rebeccaskinner @jaspervdj @ryantrinkle @romefeller @idabzo

idabzo (approved):
LGTM

romefeller (approved):

TikhonJelvis (approved):
Nice, let's do it. Thanks for gathering clear community feedback @tomjaguarpaw!

rebeccaskinner (approved):
Looks good to me, I think this is a great change. Thanks!

jaspervdj (approved):

tomjaguarpaw:
Thanks everyone for your input!

PR #126, opened 2021-08-11
Apple Silicon is now supported!

tomjaguarpaw:
As per https://github.com/haskell-infra/www.haskell.org/pull/125#discussion_r687070791

@hasufell Could you confirm this is correct?


hasufell:
> @hasufell Could you confirm this is correct?

Yes: https://gitlab.haskell.org/haskell/ghcup-hs/-/blob/master/CHANGELOG.md#01151-2021-06-11

jneira (approved):
always it is nice being able to remove warnings, thanks!

TikhonJelvis (approved):
Sweet, this is great to hear :).

jaspervdj (approved):

PR #129, opened 2021-08-13
Update downloads page

tomjaguarpaw:
Minor edits to https://github.com/haskell-infra/www.haskell.org/pull/125 of two types:

* Fix minor typos
* Replace "here" as link text with text describing the link target

I suggest that #125 be replaced with this.

### DIFF
--- a/downloads.markdown
+++ b/downloads.markdown
@@ -6,7 +6,7 @@ isDownloads: true
 
 # Downloads
 
-This page describes the installation of the Haskell toolchain, which consists of the following tools:
+The Haskell toolchain consists of the following tools:
 
 *   [GHC](https://www.haskell.org/ghc/): the Glasgow Haskell Compiler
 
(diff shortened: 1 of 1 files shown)


downloads.markdown
```diff
@@ -16,43 +16,28 @@ This page describes the installation of the Haskell toolchain, which consists of
...
-2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
-
-### Windows { #windows }
-
-1. Install GHC and cabal-install via Chocolatey
-	- [Configure Chocolatey](https://chocolatey.org/install) on your machine
-	- At an elevated command prompt, run `choco install haskell-dev`, followed by `refreshenv`.
-2. To install stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/#windows)
+1. Use [GHCup](https://www.haskell.org/ghcup/) to install
+   GHC, cabal-install, haskell-language-server and (experimentally) stack.
```
hasufell, line 22: I'd omit stack here

```suggestion
   GHC, cabal-install and haskell-language-server
```

downloads.markdown
```diff
@@ -16,43 +16,28 @@ This page describes the installation of the Haskell toolchain, which consists of
...
+2. Stack can be also installed directly as described on
```
hasufell, line 23:
```suggestion
2. Stack can be installed as described on
```

hasufell (changes requested):
I'd suggest to omit that ghcup can do stack installations for now. Experimental options don't need to be promoted. Either we stand behind it or we don't.

tomjaguarpaw:
Closing because pushed to https://github.com/haskell-infra/www.haskell.org/pull/125

PR #130, opened 2021-08-17
Add link to stack bindist

tomjaguarpaw:
Originally part of https://github.com/haskell-infra/www.haskell.org/pull/125 but logically separate so it seems better to consider it separately.


tomjaguarpaw:
To answer a question of @goldfirere (whose source I can no longer find), I don't know the reason for the particular format of the bindist links. I agree a bulleted list might be better. But we should add the stack bindist and add bulleting separately (if indeed we will do both).

TikhonJelvis (approved):

TikhonJelvis:
@tomjaguarpaw do we still want to merge this? Looks like we approved it a while ago.

tomjaguarpaw:
Yes, sure, I'll rebase it and get it merged at some point.

PR #138, opened 2021-10-29
Add Fission testimonial

expede:


rebeccaskinner (changes requested):
Thanks for the PR! It looks like the logo png needs to be resized to 300x300px. Would you mind resizing it? Otherwise this looks good to me.

tomjaguarpaw:
Currently, as you can see from [the action](https://github.com/haskell-infra/www.haskell.org/actions/runs/1397107418) it looks like the below. Is that what you mean by it needs to be 300x300, @rebeccaskinner?

screenshot

expede:
@rebeccaskinner fixed! Please let me know if you need any other changes

tomjaguarpaw:
Looks better now, thank you!

screenshot

tomjaguarpaw (approved):

rebeccaskinner (approved):

PR #143, opened 2021-11-03
call nix-build rather than 'nix build'

rebeccaskinner:


TikhonJelvis:
`nix-build` prints the logs automatically, so we can just remove that flag.

tomjaguarpaw:
Is it OK to delete this branch?

rebeccaskinner:
> Is it OK to delete this branch?

Oh yes, we can delete. This was an experiment when I was trying to fix CI the other day.

PR #144, opened 2021-11-03
fix github actions to use nix 2.4 cli

rebeccaskinner:


TikhonJelvis (approved):
Nice, looks great. I didn't know about YAML multi-line formatting—definitely makes the commands easier to read :).

I'm surprised that Nix 2.4 started requiring this experimental feature flag after not needing it in 2.3, but not much else we can do about that.

PR #145, opened 2021-11-03
fix another nix issue

rebeccaskinner:


TikhonJelvis (approved):
Looks like this should work too, but I guess we need to merge to try it :)

Issue #147, opened 2021-12-07
Make haskell-platform page redirect to the downloads page


MorrowM:
Now that the Downloads page is in decent shape I think it might be time to remove the page for the Haskell Platform and make it redirect to the downloads page. As of the time of posting haskell-platform supersedes the downloads page in the search results for "install haskell" on both Google and DuckDuckGo. I've seen a number of people struggling with their GHC/cabal installations because of this, and they end up switching to GHCup anyways, which wastes their time.

A less extreme measure would be to put a deprecation banner up, but that wouldn't be a very smooth experience for newcomers in my opinion. 

![image](https://user-images.githubusercontent.com/69002988/145032245-1ba45c12-646a-4422-9cff-2a3766bc2e6e.png)

tomjaguarpaw:
Thanks very much for reporting this. I agree. There is some related discussion in https://github.com/haskell-infra/www.haskell.org/issues/136#issuecomment-944288891

tomjaguarpaw:
Fixing that will close this issue too: https://github.com/haskell-infra/haskell-admins/issues/4

tomjaguarpaw:
Fixed by https://github.com/haskell-infra/www.haskell.org/pull/149#issuecomment-997019362

PR #149, opened 2021-12-12
Add redirect from Haskell Platform page to Downloads page

MorrowM:
#147 seems to have been gotten good support, so here's the PR with the change. 

I couldn't find any good documentation indicating whether anything needs to be changed in the haskell-platform repository, so feel free to tell me if anything else needs to be done. Or, of course, if the redirect should be done differently/not via Hakyll.

### DIFF
--- a/builder/site.hs
+++ b/builder/site.hs
@@ -1,6 +1,7 @@
 --------------------------------------------------------------------------------
-{-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TupleSections     #-}
+{-# LANGUAGE TypeApplications  #-}
 import           Data.Aeson
 import qualified Data.ByteString.Lazy          as BL
 import           Data.Monoid                   ((<>))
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Thanks for this. You are right to point out the [haskell-platform](https://github.com/haskell/haskell-platform) repository as the source of the `platform/` subsite. According to the [work-in-progress PR to list subsites](https://github.com/haskell-infra/www.haskell.org/pull/103) we don't know the owner of the `platform/` subsite. I suspect it's simply the haskell.org committee.

If we are going to remove mention of the Platform from haskell.org (which I support) then we will need to add redirects for all the following pages:

* https://www.haskell.org/platform/index.html
* https://www.haskell.org/platform/mac.html
* https://www.haskell.org/platform/linux.html
* https://www.haskell.org/platform/windows.html
* https://www.haskell.org/platform/prior.html
* https://www.haskell.org/platform/download.html
* https://www.haskell.org/platform/contents.html

and the server admins should delete the files for the `platform/` subsite that are currently present on the server.

MorrowM:
I've added redirects for the pages you've listed.

tomjaguarpaw:
Thank you! Hopefully other committee members will comment in due course.

tomjaguarpaw (approved):

tomjaguarpaw:
The committee approved this change at today's meeting. Hopefully I will get round to it at the weekend.

gbaz:
Since I was doing the redirects for haddock, I've added the redirects to the nginx config and they're now in place for this as well.

tomjaguarpaw:
Great, thank you!

PR #153, opened 2022-01-03
More robust command to find haskell-org-site executable path

tomjaguarpaw:


buildAndWatch
```diff
@@ -27,9 +27,8 @@ function buildAndWatchWithoutNix() {
 
     pushd builder
     cabal v2-build
-    builderPath=$(find dist-newstyle -name 'haskell-org-site' -type f)
+    builder=$(cabal v2-exec -- which haskell-org-site)
     popd
-    builder="./builder/${builderPath}"
```
tomjaguarpaw, line 32: N.B. the path returned by the new `cabal v2-exec` line is absolute. No need to prefix it with anything any more.

rebeccaskinner (approved):
This is much better- thanks for the PR!

PR #154, opened 2022-01-03
Move site files to site/ subdirectory

tomjaguarpaw:
It's rather messy otherwise, and in particular the `**/*.markdown` rule matches `.markdown` files belonging to the source of dependent packages in in the dist-newstyle directory, preventing the site build from completing successfully when using cabal.

Does this make sense to you, @rebeccaskinner? I think you're probably the person who knows most about the build process.


rebeccaskinner (approved):
This looks good to me, and definitely improves the overall organization of the repo.

tomjaguarpaw:
(Deployment failed after this PR(1) perhaps because another deployment from a different PR(2) was running at the same time. Looks like the end state is fine though. If we run across this too often we might want to consider some sort of locking.)

(1): https://github.com/haskell-infra/www.haskell.org/runs/4700066521?check_suite_focus=true

(2): https://github.com/haskell-infra/www.haskell.org/runs/4700068576?check_suite_focus=true

PR #161, opened 2022-01-22
Add lhbg to list of introductory books

soupi:
Can I add [lhbg](https://lhbg-book.link) to the list of intro books under the documentation page?

### DIFF
--- a/site/documentation.markdown
+++ b/site/documentation.markdown
@@ -22,6 +22,7 @@ If you are new to Haskell and are not sure where to start from, we recommend [CI
 *   \[$$\] [Haskell: The Craft of Functional Programming](http://www.haskellcraft.com/craft3e/Home.html)
 *   \[$$\] [The Haskell School of Music](http://euterpea.com/haskell-school-of-music/)
 *   \[$$\] [Get Programming with Haskell](https://www.manning.com/books/get-programming-with-haskell)
+*   [Learn Haskell by building a blog generator](https://lhbg-book.link)
 
 ## Intermediate Haskell Books
 


tomjaguarpaw:
This looks great, thanks!

tomjaguarpaw:
FYI this will require at least one more committee review.

site/documentation.markdown
```diff
@@ -22,6 +22,7 @@ If you are new to Haskell and are not sure where to start from, we recommend [CI
 *   \[$$\] [Haskell: The Craft of Functional Programming](http://www.haskellcraft.com/craft3e/Home.html)
 *   \[$$\] [The Haskell School of Music](http://euterpea.com/haskell-school-of-music/)
 *   \[$$\] [Get Programming with Haskell](https://www.manning.com/books/get-programming-with-haskell)
+*   [Learn Haskell by building a blog generator](https://lhbg-book.link)
```
TikhonJelvis, line 25: Looks like a great addition to me :).

Is the capitalization of the title intentional? I'd normally expect title case for a book (ie "Learn Haskell by Building a Blog Generator"), but it's perfectly fine to do something different if it's a stylistic choice or something.

TikhonJelvis (approved):

site/documentation.markdown
soupi: It is I guess, I don't know why I'm not a huge fan of capitalizing all words in a title 😅, but I don't mind changing that to fit the rest of the list.
FWIW the capitalization is similar to how 'Haskell Programming from first principles' is capitalized in the list.

TikhonJelvis: Since that's what you've chosen for the book itself, makes sense to keep it the same on the resources page too. I just wanted to double-check that this was what you intended :).

soupi:
Thanks!

PR #163, opened 2022-01-31
Specify which systems the installation instructions support

tomjaguarpaw:
I'm not 100% sure Stack supports FreeBSD so I asked them: https://github.com/commercialhaskell/stack/issues/5674. Even if it doesn't I think it's better to merge this and adjust the Stack info later.


jaspervdj (approved):

hasufell:
The stack installer script doesn't work on FreeBSD.

Issue #165, opened 2022-01-31
Euterpea website links to www.haskell.org/platform


tomjaguarpaw:
https://www.euterpea.com/

This is particularly confusing because `www.haskell.org/platform` doesn't redirect to the downloads page (although with a trailing slash it does).

Mitigations

* Add back a platform page that links to download page (see https://github.com/haskell-infra/www.haskell.org/issues/164)
* Ask Euterpea website to update their instructions

tomjaguarpaw:
Next action: Ask Euterpea website to update their instructions

georgefst:
I mentioned this to the author in https://github.com/Euterpea/Euterpea2/pull/52, and references to the Haskell Platform do seem to have been fully removed from the website (alas, I think it's the only suggestion from that thread which was actually acted on).

tomjaguarpaw:
Nice, thanks!

Issue #164, opened 2022-01-31
Lingering references to Haskell Platform are confusing


tomjaguarpaw:
There are lingering references around the Haskell ecosystem to Haskell Platform, which is deprecated and mention of it removed from the site.  For example, [a user report](https://www.reddit.com/r/learnprogramming/comments/sgvk35/is_haskell_platform_no_longer_supported/).

Perhaps we can add back a `platform/` page which simply confirms that HP is deprecated and guides users to the downloads page.

(If we can also find old resources that mention the platform and ask them to update themselves then so much the better.)

ulysses4ever:
One option is to add back the platform page but make it redirect immediately via the [meta refresh](https://en.m.wikipedia.org/wiki/Meta_refresh) method instead of trying to explain anything on it. I think that'd be more straightforward.

tomjaguarpaw:
Unfortunately I don't think it's sufficient to *just* redirect the downloads page. Users will quite understandably confused about why all references to Haskell Platform on haskell.org just disappeared, despite claims in the books that they read that it is the official way to get Haskell. A simple note that Haskell Platform has been replaced by the method of the downloads page would suffice.

tomjaguarpaw:
Closing this as it has been addressed by https://github.com/haskell-infra/www.haskell.org/pull/167, but we should collect a list of resources that mention Haskell Platform and ask them to update themselves.

PR #167, opened 2022-02-01
Add platform/ page describing that the Haskell Platform is deprecated

tomjaguarpaw:
and link to the Downloads page instead

Part of:

* https://github.com/haskell-infra/www.haskell.org/issues/164
* https://github.com/haskell-infra/www.haskell.org/issues/165

### DIFF
--- a/site/platform.markdown
+++ b/site/platform.markdown
@@ -1,0 +1,13 @@
+---
+title: Haskell Platform
+page: platform
+isDownloads: true
+---
+
+# Haskell Platform
+
+## The Haskell Platform is deprecated
+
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
This is extremely bare bones, but given that there are lingering references to <http://www.haskell.org/platform> all around the web (and in printed books) I think we should at least get *something* up there, before we decide what to do long term.

screenshot

tomjaguarpaw:
This is deployed but looks like the redirect from www.haskell.org/platform to www.haskell.org/downloads takes precedence.  I have asked on #haskell-infrastructure to have the redirect removed.

PR #171, opened 2022-02-08
Note that Stack does not support FreeBSD

tomjaguarpaw:
It seems that [Stack doesn't support FreeBSD](https://github.com/commercialhaskell/stack/issues/5674) so we should probably note that fact.

### What it looks like now

screenshot


jaspervdj (commented):
Minor nitpick: can we move this note inside the (2) list item?

tomjaguarpaw:
Sure, how about this?

screenshot

tomjaguarpaw:
Any more thoughts @jaspervdj?

jaspervdj (approved):
LGTM!

Issue #140, opened 2021-11-01
IOHK's logo wrong size?


tomjaguarpaw:
Looks like IOHK's logo might be the wrong size.

![screenshot](https://user-images.githubusercontent.com/1951567/139668553-1f27f1d1-a913-467e-ad3b-b7fca6e54155.png)

tomjaguarpaw:
Fixed by https://github.com/haskell-infra/www.haskell.org/pull/173

PR #173, opened 2022-02-09
Fix oddly sized logos in testimonials

sekunho:
Read #140 and found out there were some others that could be improved.

IOHK before:

image

IOHK after:

image

e-bot7 before:

image

e-bot7 after:

image

Fission before:

image

Fission after:

image


tomjaguarpaw (approved):

tomjaguarpaw:
Fantastic, thank you! Perhaps we want to keep the SVG though, in case we need it in the future?

sekunho:
Alright, I'll add that back in.

rebeccaskinner (commented):
I noticed this changes the IOHK logo from a version with text to one without. Is this an official logo they provide? I think the changes look great, and I'm not worried about the resizing on the others since we're not changing the logo, but I don't want to change a companies branding without their permission.

tomjaguarpaw:
In that case, @sekunho can you remove the IOHK change from this PR and make a separate PR for it? We can get this one in whilst we seek confirmation from IOHK.

sekunho:
@rebeccaskinner @tomjaguarpaw Yeah, I detailed this in the commit message https://github.com/haskell-infra/www.haskell.org/pull/173/commits/de3fb38ab86a7509dcf14b8fdafe2f154d695d0f.

> Merely resizing it to fit nicely in a 300x300 resolution wasn't enough
since the logo would be too small, give how wide it is. So I got one
without the text from IOHK's official media kit.
>
> https://iohk.io/en/media-kit/

It's their official logo.

tomjaguarpaw:
OK then, this looks fine. @rebeccaskinner, do you agree?

rebeccaskinner (approved):
This is looks great to me, thanks for confirming that we're good to use this IOHK logo!

rebeccaskinner:
> OK then, this looks fine. @rebeccaskinner, do you agree?

yep, approved!

sekunho:
Thanks for everyone's time!

tomjaguarpaw:
Thank you for your contribution, @sekunho!

PR #180, opened 2022-02-17
Add Effective Haskell to Intermediate Books section

tomjaguarpaw:


rebeccaskinner:
I'm not going to approve this PR since it would be a conflict of interest, but I do appreciate the interest in sharing my book.

I think we could put the book in the beginners section, since it doesn't require any prior familiarity with Haskell.

I also wonder whether or not we should note that the book is currently in Beta, and so it might still contain some minor errors and isn't yet feature complete. That is called out on the store page, and I think the current state of the book should still mean that it's useful for most learners, but I thought I'd mention it.

site/documentation.markdown
```diff
@@ -27,6 +27,7 @@ If you are new to Haskell and are not sure where to start from, we recommend [CI
 ## Intermediate Haskell Books
 
 *   [Developing Web Applications with Haskell and Yesod](http://www.yesodweb.com/book)
+*   \[$$\] [Effective Haskell](https://www.pragprog.com/titles/rshaskell/effective-haskell/)
```
TikhonJelvis, line 30: Looks great to me. We could make the beta status of the book pretty simply—should be fine as long as we don't forget to update it when it's out of beta!

I reviewed the book in detail earlier and would definitely agree that it could be moved to the beginner section. It covers some intermediate material, but it does not assume any Haskell knowledge and starts by teaching basic Haskell ideas suitable for people learning the language for the first time.

```suggestion
*   \[$$\] [Effective Haskell (beta)](https://www.pragprog.com/titles/rshaskell/effective-haskell/)
```

TikhonJelvis (approved):

tomjaguarpaw:
OK, I moved it to the beginners section and marked it as beta.

PR #182, opened 2022-02-20
Add link to differences between GHC and the language report

noughtmare:
I think it is important for readers of the report to know how GHC differs.

### DIFF
--- a/site/documentation.markdown
+++ b/site/documentation.markdown
@@ -71,7 +71,7 @@ Curated resources put together by Haskellers:
 
 Manuals and guides that cover common Haskell tooling:
 
-*   [GHC User Guide](http://www.haskell.org/ghc/docs/latest/html/users_guide/)
+*   [GHC User's Guide](http://www.haskell.org/ghc/docs/latest/html/users_guide/)
 *   [Cabal Homepage And Quick Links](https://www.haskell.org/cabal/)
 *   [Cabal User Guide](http://www.haskell.org/cabal/users-guide/)
 *   [Stack User Guide](https://docs.haskellstack.org/)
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Sounds sensible. Could you please move the link to be around "the GHC User's Guide" (and match its punctuation, i.e. "User's" not "User")?  I can see the earlier links use "here" as anchor text, but that's pretty bad practice and I don't want to perpetuate it.

noughtmare:
Hmm... The manuals and guides list includes "GHC User Guide" should that be changed too?

tomjaguarpaw:
Yes, well spotted. Feel free to add that to this PR.

noughtmare:
I see now that it also lists `Cabal User Guide` , `Stack User Guide`, and `Haddock User Guide`. Can I fix those too in this pull request? It's unrelated, but it feels awkward to leave those in.

Also, the Stack User Guide does call itself "User guide": https://docs.haskellstack.org/en/stable/GUIDE/

tomjaguarpaw:
If you want to fix more then please go ahead! However, the text we use should be the same as the way that the guides refer to themselves. Cabal and Stack both seem to use "User Guide" so we should stick to that (despite not actually linking directly to what Stack calls its "User Guide").

noughtmare:
Oh, the link of the cabal guide is www.haskell.org/cabal/users-guide/, so I assumed it would be User's Guide... I'll just change the GHC one then.

tomjaguarpaw:
Ah I see. That link is outdated anyway. It should point to <https://cabal.readthedocs.io/en/3.6/>.

tomjaguarpaw:
Very nice, thank you!

tomjaguarpaw (approved):

tomjaguarpaw:
We'll need to wait for another committee member to approve too.

noughtmare:
> Ah I see. That link is outdated anyway. It should point to https://cabal.readthedocs.io/en/3.6/.

I think the haskell.org link redirects to that, which is nice because the haskell.org link should be stable even if cabal releases another version or decides to move away from readthedocs.

tomjaguarpaw:
Yes, OK, that seems like a good reason to preserve it.

TikhonJelvis (approved):
Looks good. Thanks for the PR :)

PR #184, opened 2022-03-05
Remove outdated irc log link

robx:
The logs don't seem to have survived the move to libera.


tomjaguarpaw:
True, but the older logs still exist and are presumably interesting. I would prefer keeping them around. I would even more strongly prefer adding a new source of logs, if anyone knows one.

robx:
I've updated the PR to keep the link but mention that it's not kept up-to-date.

tomjaguarpaw (approved):

tomjaguarpaw:
OK, that works for me. We'll need to wait for one more committee member to approve this before merging.

TikhonJelvis (approved):

TikhonJelvis:
Thanks for the fix! Sorry it took a while to merge in.

Issue #170, opened 2022-02-08
"Owners" needed for the "alternative installation options"


tomjaguarpaw:
It would be great to have "owners" for each of the "[alternative installation options](https://www.haskell.org/downloads/)" (listed below). That is, for each alternative installation option there should be someone who is responsible for making sure that installation option is specified clearly, works, and is kept up to date. If we can't find an owner for an installation option then I suggest we remove that installation option because (a) lack of owner probably implies the option is not very popular anyway and (b) the information is likely to get outdated and broken, causing more harm than good.

If anyone has any thoughts then please comment here. Unless the discussion here suggests otherwise, after a week or two I'll add tickets for each installation option, to find an owner or remove the option. Once we have those tickets we can post widely on the usual Haskell channels to find volunteer owners.

### Current alternative installation options

* Linux Ubuntu (confusing (see https://github.com/haskell-infra/www.haskell.org/issues/16) and probably outdated)
* Linux Debian (links to https://downloads.haskell.org/~debian/ which doesn't support Debian 11 Bullseye)
* Linux Fedora
* Linux EPEL for RHEL/CentOS/etc
* Linux Arch
* Linux openSUSE Leap
* Linux openSUSE Tumbleweed
* Linux Gentoo
* Windows Chocolatey

tomjaguarpaw:
It may also be useful to bear in mind a ranking of most popular Linux distributions, such as [the ranking from DistroWatch](https://distrowatch.com/dwres.php?resource=popularity).

jaspervdj:
We discussed this during the Haskell.org meeting and we would prefer to significantly shorten the alternative installation options, or remove it entirely.

tomjaguarpaw:
OK, so if anyone has any strong feelings about continuing to present "alternative installation options" please speak up, especially if you can help us support them with your expertise!

I will also circulate this on Haskell Discourse at some point to widen the debate.

tomjaguarpaw:
Posted to

* [Discourse](https://discourse.haskell.org/t/rfc-removing-alternative-installation-methods-from-haskell-org-or-finding-them-owners/4318)
* [Reddit](https://www.reddit.com/r/haskell/comments/tuq21i/rfc_removing_alternative_installation_methods/)
* [Twitter](https://nitter.net/tomjaguarpaw/status/1510327991729111041)
* [Haskell Cafe](https://mail.haskell.org/pipermail/haskell-cafe/2022-April/135185.html)

[...]

PR #194, opened 2022-04-10
Remove distro-specific instructions and only leave links

tomjaguarpaw:
The links will go out-of-date less quickly so will be easier to maintain.

The original patch was by @hasufell. Some changes were made by @tomjaguarpaw.

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -24,10 +24,7 @@ This page describes the installation of the Haskell toolchain, which consists of
 
 ### Via native OS package manager
 
-Alternatively, many operating systems provide GHC, cabal and Stack
-through their native package manager.  The packages are often
-out-of-date but if you prefer to use this method of installation then
-you will find useful links below.
+Alternatively, many operating systems provide GHC, cabal and Stack through their native package manager.  The packages are often out-of-date but if you prefer to use this method of installation then you will find useful links below.
 
 <p><a data-toggle="collapse" href="#collapse-linux" class="btn btn-xs btn-primary">Show Linux distros</a></p>
 
@@ -72,10 +69,7 @@ you will find useful links below.
 
 #### Arch
 
-**Do not use the Haskell development tools provided by Arch, they are
-broken.** For more information see
-[[1]](https://dixonary.co.uk/blog/haskell/cabal-2020)
-[[2]](https://stackoverflow.com/questions/65643699/what-is-the-suggested-way-of-setting-up-haskell-on-archlinux/65644318#65644318).
+**Do not use the Haskell development tools provided by Arch, they are broken.** For more information see [[1]](https://dixonary.co.uk/blog/haskell/cabal-2020) [[2]](https://stackoverflow.com/questions/65643699/what-is-the-suggested-way-of-setting-up-haskell-on-archlinux/65644318#65644318).
 
 </div>
 
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -9,168 +9,92 @@ isDownloads: true
 This page describes the installation of the Haskell toolchain, which consists of the following tools:
 
 *   [GHC](https://www.haskell.org/ghc/): the Glasgow Haskell Compiler
-
 *   [cabal-install](https://cabal.readthedocs.io): the Cabal installation tool for managing Haskell software
-
 *   [Stack](https://docs.haskellstack.org): a cross-platform program for developing Haskell projects
-
 *   [haskell-language-server](https://github.com/haskell/haskell-language-server) (optional): A language server for developers to integrate with their editor/IDE
 
-## Installation instructions { #ghcup }
+## Recommended installation instructions
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
 
 1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
-2. To install Stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/) *(N.B. Stack does not support FreeBSD)*
+2. To install Stack, follow the [Stack installation guide](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
 
 * * *
 
-## Alternative installation options { #ghc-install-manual }
+### Via native OS package manager
 
-*   [Using a package manager](#package-manager)
-
-*   [Official bindists](#bindists)
-
-*   [Building from source](#from-source)
-
-*   [Other options](#other-options)
-
-### Using a package manager { #package-manager }
-
-#### Linux
-
-Refer to your distribution package manager documentation. For convenience, below are a few distribution specific instructions, outlining 3rd party repository use as well.
-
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
I decided to omit mention of Nix here. @maralorn [has agreed to be "owner" of the Nix installation method](https://github.com/haskell-infra/www.haskell.org/issues/170#issuecomment-1092021854) so he can submit a PR on top with the Nix content that he wishes to maintain.

tomjaguarpaw:
Please see https://github.com/haskell-infra/www.haskell.org/issues/170#issuecomment-1094336738 for the feedback that was taken into account to produce this PR.

site/downloads.markdown
```diff
@@ -9,168 +9,92 @@ isDownloads: true
...
-
-While the Haskell toolchain is available in the main gentoo repository, it is recommended
-to use the more up-to-date [gentoo-haskell](https://github.com/gentoo-haskell/gentoo-haskell) overlay. This can be done using layman:
-
-```
-sudo layman -a haskell
-sudo emerge --ask dev-lang/ghc dev-haskell/cabal-install
-```
+**Do not use the Haskell development tools provided by Arch, they are
+broken.** For more information see
```
TikhonJelvis, line 76:
```suggestion
**Haskell tools and libraries natively packaged by Arch do not work for local Haskell development.** For more information see
```

site/downloads.markdown
```diff
@@ -9,168 +9,92 @@ isDownloads: true
...
-
-Refer to your distribution package manager documentation. For convenience, below are a few distribution specific instructions, outlining 3rd party repository use as well.
-
-*Note: Most Linux distros don't allow different versions of GHC to be installed in parallel. Sometimes the GHC packages are a bit outdated as well. In either case, consider using [GHCup](#ghcup) instead.*
-
-*Note: haskell-language-server isn't packaged by most distributions. Instead you will have to install it manually, see the [release page](https://github.com/haskell/haskell-language-server/releases), unless you use VSCode, which bootstraps everything automatically. This is optional.*
+Alternatively, many operating systems provide GHC, cabal and Stack
+through their native package manager.  The packages are often
+out-of-date but if you prefer to use this method of installation then
+you will find useful links below.
```
TikhonJelvis, line 30: I'd lean towards a slightly different tone here:

```suggestion
Many operating systems provide GHC, cabal and Stack through their package managers. Note that OS packages can often be out of date.
```

site/downloads.markdown
TikhonJelvis: Seems like we could be a bit more specific here—if I'm understanding the situation correctly, the Arch approach is fine for stuff like packaging Pandoc for Arch users, but it does not work well for *doing Haskell development in general*.

TikhonJelvis (approved):
Looking good overall. Most of the changes have already had lots of discussion, so the core choices in this PR all make sense.

I left a couple of suggested change to wording, but they're pretty minor—definitely not blockers to get this merged.

Also, as a minor style thing, it's nice to keep Markdown paragraphs on one line. This lets people visually wrap Markdown documents in their editor to whatever width they need while also keeping Git diffs cleaner, especially when different editors hard-wrap in different ways.

site/downloads.markdown
```diff
@@ -9,168 +9,92 @@ isDownloads: true
...
 *   [GHC](https://www.haskell.org/ghc/): the Glasgow Haskell Compiler
-
 *   [cabal-install](https://cabal.readthedocs.io): the Cabal installation tool for managing Haskell software
-
 *   [Stack](https://docs.haskellstack.org): a cross-platform program for developing Haskell projects
-
 *   [haskell-language-server](https://github.com/haskell/haskell-language-server) (optional): A language server for developers to integrate with their editor/IDE
 
-## Installation instructions { #ghcup }
+## Recommended installation instructions
```
rebeccaskinner, line 16: I wonder if we might not want to add something like `(All Operating Systems)` at the end here so someone glancing at the document will start here. I guess perhaps the `recommended` covers it, but I'd like to draw eyes to this section rather than the distro packages, since I think on average people will have a better experience avoiding the distro packages.

site/downloads.markdown
```diff
@@ -9,168 +9,92 @@ isDownloads: true
...
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
 
 1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
-2. To install Stack, follow the instructions [here](https://docs.haskellstack.org/en/stable/install_and_upgrade/) *(N.B. Stack does not support FreeBSD)*
+2. To install Stack, follow the [Stack installation guide](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
```
rebeccaskinner, line 21: Would it make sense for these to be bullet points rather than a numbered list? The numbered list implies to me that you need to follow each step in order, rather than having a choice of install method.

site/downloads.markdown
```diff
@@ -9,168 +9,92 @@ isDownloads: true
...
 *   [cabal-install](https://cabal.readthedocs.io): the Cabal installation tool for managing Haskell software
-
 *   [Stack](https://docs.haskellstack.org): a cross-platform program for developing Haskell projects
-
 *   [haskell-language-server](https://github.com/haskell/haskell-language-server) (optional): A language server for developers to integrate with their editor/IDE
 
-## Installation instructions { #ghcup }
+## Recommended installation instructions
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
```
rebeccaskinner, line 18: Nit-pick, but I wonder if we might want to remove this line altogether in light of the other changes?

tomjaguarpaw:
> as a minor style thing, it's nice to keep Markdown paragraphs on one line

OK, I undid the hard wrapping.

rebeccaskinner (approved):
I had a couple of thoughts but I'm happy to approve as-is since I think this is still an improvement.

site/downloads.markdown
tomjaguarpaw: Sure, but someone packaging Pandoc for Arch users already knows that, right?  The ratio among visitors to this page of  Arch users who are not trying to package for Arch : Arch developers who are trying to package for arch must be something like 1000 : 1, or more right?  In which case I think it's probably unhelpful to be more nuanced, even if it is technically correct.

Put another way, the people who would be negatively effected by the lack of nuance in the original wording already know the solution anyway, so I don't think there's much point going out of our way to accommodate them at the expense of presenting a clear message.

site/downloads.markdown
tomjaguarpaw: Yes, that's a good point. However, the numbered list predates this PR so I suggest we leave that for a separate PR.

site/downloads.markdown
hasufell: We want to make *very* clear to users that they shouldn't use arch packages. The word **broken** is key here.

TikhonJelvis: Seems fine either way.

jaspervdj (approved):

jaspervdj:
Thanks for spearheading this @tomjaguarpaw!  I agree with Rebecca's small remarks, but overall looks good to me.

romefeller (approved):

PR #198, opened 2022-04-24
bump hakyll upper bound to <4.16

mhitza:
When trying to build/run the site locally with GHC 9.2.2 I needed to change the hakyll upper bound. I could have gone incrementally to 4.15.x to find a version that could still build, but found it easier to bump it all the way to 4.16. Haven't seen any breaking changes in the hakyll changelog in the 4.14 -> 4.15.latest changelist.


tomjaguarpaw (approved):

tomjaguarpaw:
Looks sensible to me. We'll need to wait for another committee approval.

PR #200, opened 2022-05-21
downloads: Add entry for NixOS 

maralorn:
As discussed in https://github.com/haskell-infra/www.haskell.org/issues/170#issuecomment-1092032433

I have appended the NixOS entry to be polite. I wonder if there could be a more principled approach to ordering the entries.
At the moment I guess it’s "felt importance"? Alphabetically might be an improvement. Sadly the Haskell survey does not ask for the distributions that people primarily use.

Note: It is our aim to improve and move the documentation. Still the current documentation is a lot better than nothing. So I think we should put the link here for the while. I will update it as soon as we made progress on this.

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -70,6 +70,22 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 #### Arch
 
 **Do not use the Haskell development tools provided by Arch, they are broken.** For more information see [[1]](https://dixonary.co.uk/blog/haskell/cabal-2020) [[2]](https://stackoverflow.com/questions/65643699/what-is-the-suggested-way-of-setting-up-haskell-on-archlinux/65644318#65644318).
+</div>
+
+<p><a data-toggle="collapse" href="#collapse-nixos" class="btn btn-xs btn-primary">Show Nix/NixOS instructions</a></p>
+
+<div id="collapse-nixos" class="collapse">
+<!-- This installation method is owned by @maralorn -->
+
+*for Nix the package manager on the distribution NixOS and other OS*
+
+* **Using ghcup on NixOS does not work.** Use the official packages instead.
+* You can get started by installing
+    [GHC](https://search.nixos.org/packages?show=ghc&type=packages&query=ghc),
+    [Haskell Language Server](https://search.nixos.org/packages?show=haskell-language-server&type=packages&query=haskell-language-server) and
+    [cabal-install](https://search.nixos.org/packages?show=cabal-install&type=packages&query=cabal-install) (or
+    [Stack](https://search.nixos.org/packages?show=stack&type=packages&query=stack)) via Nix.
+* Read the [documentation](https://haskell4nix.readthedocs.io/nixpkgs-users-guide.html#how-to-create-a-development-environment) to learn more about development environments and packaging for Haskell with Nix.
 
 </div>
 


site/downloads.markdown
```diff
@@ -71,6 +71,18 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 
 **Do not use the Haskell development tools provided by Arch, they are broken.** For more information see [[1]](https://dixonary.co.uk/blog/haskell/cabal-2020) [[2]](https://stackoverflow.com/questions/65643699/what-is-the-suggested-way-of-setting-up-haskell-on-archlinux/65644318#65644318).
 
+#### NixOS
+
+[](This installation method is owned by @maralorn)
+
+* [Documentation for Haskell development and packaging with nix](https://haskell4nix.readthedocs.io/nixpkgs-users-guide.html#how-to-create-a-development-environment)
```
tomjaguarpaw, line 78: Should "Nix" be capitalized? It seems so, [according to Wikipedia](https://en.wikipedia.org/wiki/Nix_(package_manager)).

```suggestion
* [Documentation for Haskell development and packaging with Nix](https://haskell4nix.readthedocs.io/nixpkgs-users-guide.html#how-to-create-a-development-environment)
```

tomjaguarpaw:
Thanks! I'm not a Nix user, but as I understand it, Nix and NixOS are slightly different. Are you sure you're happy having NixOS as a subentry under Linux or should it belong at the top level? (I don't mind either way)

site/downloads.markdown
```diff
@@ -71,6 +71,18 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 
 **Do not use the Haskell development tools provided by Arch, they are broken.** For more information see [[1]](https://dixonary.co.uk/blog/haskell/cabal-2020) [[2]](https://stackoverflow.com/questions/65643699/what-is-the-suggested-way-of-setting-up-haskell-on-archlinux/65644318#65644318).
 
+#### NixOS
+
+[](This installation method is owned by @maralorn)
```
tomjaguarpaw, line 76: Regrettably it turns out that this trick for "Markdown comments" doesn't play well with our "link checker" than checks all links exist. The "pseudo link" `This installation method is owned by @maralorn` doesn't exist, so link checker fails. I think we should probably switch to:

```suggestion
<!-- This installation method is owned by @maralorn -->
```

maralorn:
> Thanks! I'm not a Nix user, but as I understand it, Nix and NixOS are slightly different. Are you sure you're happy having NixOS as a subentry under Linux or should it belong at the top level? (I don't mind either way)

I have considered this and given this a second attempt.

> Is nix still "jailbreak"ing packages and removing their version bound requirements? I know I've had data loss because they "didn't believe me" that I required a newer version of 'directory' in the past. If they're still doing that at all I'd want a very large warning.

Yes, we still do jailbreaking, mostly because devs don‘t keep up with relaxing their bounds. Your story sounds terrible and is a good reminder to be careful with this. I cannot promise that this has actually become better, because I have not been in the business very long. But these days we are careful about this. In general when bounds seem weird, especially lower bounds, we check changelogs for explanations/breaking changes and contact upstream about it. We have policies helping with this, which we enforce in code review. At the very least I can say that Nix is relatively wide spread in the Haskell community and I haven‘t seen any complaints about issues like that in the last years.

I would not like to have a warning like that.

tomjaguarpaw (approved):
Thanks! Looks fine to me. Would be great to have a review by someone familiar with Nix (@TikhonJelvis, @rebeccaskinner?).

TikhonJelvis (approved):

Issue #166, opened 2022-01-31
Further simplification of downloads page


Bodigrim:
https://github.com/haskell-infra/www.haskell.org/blob/bcbc685b53bdd93ca4cbd64ab8ec5be35747475a/site/downloads.markdown#L9-L17

I suggest we remove this text: I cannot imagine what kind of purpose it serves nowadays. If I were a novice, I'd be uttterly confused. 

* Is Glasgow Haskell Compiler a right tool for me? How is it different from other Haskell compilers? Are there Edinburgh HC or Aberdeen HC?
* "cabal-install is the Cabal installation tool" - ah right, very helpful, thanks, much clearer ;) Again, is it a right tool for me? Do I need to manage Haskell software whatever it means? Can I use Cabal to develop Haskell projects?
* "stack is a cross-platform program" is another truism, which is not even true (today `stack` supports less platforms than `cabal` because of Mac M1). Can I use it to manage Haskell software, which is seemingly opposed to "develop"?
* We have just defined what consists a Haskell toolchain, but it appears that not all installation options below cover all of it. Then what was the point of such detailed definition?

I suggest we scrap all of this and free space for two large buttons to download `ghcup` and `stack` respectively. If this line of thinking is acceptable, I'm happy to prepare a PR.

TikhonJelvis:
That sounds like a better way of presenting the information to me. I would expect that almost nobody who would use the downloads page—mostly newer Haskellers—would ever need to go to the GHC website or install GHC manually.

Looks like that's the only part of the page that describes what `haskell-language-server` is. Presenting it in parallel with Stack/ghcup is confusing, but it's also something that seems worth highlighting even for new Haskellers. How should we present it on the downloads page?

Bodigrim:
`ghcup` asks whether to install `haskell-language-server` during bootstrap process, which is IMO a reasonable visibility.

gbaz:
I'm not sure this simplification is helpful -- people need to know that a bunch of different tools end up on their system, and have some idea of what they are -- there's a compiler, there's a build tool (or two!), etc. I'm fine with larger highlighted buttons, but if the descriptions are bad, they should be fixed. At some point, someone will want to know the difference between ghc, stack, and cabal, and why all three are on their system.

Bodigrim:
[...]

PR #201, opened 2022-05-22
downloads: make it clear that "recommended ways" are alternatives

ulysses4ever:
The current wording ("## Recommended installation instructions ... 1... 2...") makes me think that I should perform both steps (1 and 2), while in fact they are alternatives.

I remember someone brought up this concern before, during one of the later restructurings of Downloads, but I can't remember who and where. So, there are at least two of us who are confused by the current version.

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -17,8 +17,10 @@ This page describes the installation of the Haskell toolchain, which consists of
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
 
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Thanks for your interest in this. My understanding, though, was that we _did_ want new users to perform both steps. That way they are set up to follow any subsequent tutorial or source of instructions, regardless of whether it uses cabal or stack. It's worth discussing this more though. We need a more coherent plan for what this page should look like. There are some tickets touching on this already, for example https://github.com/haskell-infra/www.haskell.org/issues/166

Issue #189, opened 2022-04-03
Updating the Documentation page


reubenharry:
This issue is somewhat related to #61 and #136 and the more general goal of making the Haskell "on-boarding" experience smooth.

Below is a suggested new version of the current Documentation page (https://www.haskell.org/documentation/). I don't have any web design skills, so am just presenting it as Markdown, which is not to say that a fancier layout wouldn't look better. The general philosophy is to present fewer options, so that there's a clearer, more centrally endorsed pathway of what to look at, and less chances of broken links. The equivalent Rust page (https://www.rust-lang.org/learn) seems to me like something to aim for (Julia's one - https://julialang.org/learning/ - feels a bit cluttered IMO).

Beyond updating the page, I think it should also be more prominently linked from the front page (maybe just a larger link - see Rust), and perhaps named "Learn", like Rust's, instead of the drier "Documentation".  

I'm also assuming that the Get Started instructions would be elsewhere (i.e. a link to GHCUP on the front page).

This is intended as a draft, not a complete version. The idea is to get a sense of whether people think this is
a) generally a good idea, but disagree with the specifics
b) bad idea, makes things worse, but thanks for the suggestions!
c) good idea, and here are some further suggestions

-------

# Resources for Newcomers 

New to Haskell and want to learn the basics of the world's leading pure, lazy and statically typed language? You're in the right place!

## Learning the language

[Learn You a Haskell](http://learnyouahaskell.com/) is a community favorite book. There's also a [notebook version](https://github.com/jamesdbrock/learn-you-a-haskell-notebook) if you like to learn interactively.

<details>
  <summary>Other intro Haskell books (click to expand)</summary>
  
*   \[$$\] [Haskell from the Very Beginning](https://www.haskellfromtheverybeginning.com/)
*   [Learn You a Haskell for Great Good!](http://learnyouahaskell.com/)
*   [Real World Haskell](http://book.realworldhaskell.org/)
*   \[$$\] [Haskell Programming from first principles](http://haskellbook.com)
*   \[$$\] [Beginning Haskell](http://www.apress.com/9781430262503)
*   \[$$\] [Thinking Functionally with Haskell](http://www.cambridge.org/us/academic/subjects/computer-science/programming-languages-and-applied-logic/thinking-functionally-haskell)
*   \[$$\] [Programming in Haskell](http://www.cs.nott.ac.uk/~pszgmh/pih.html)
[...]

Issue #136, opened 2021-10-11
Optimizing initial experience on haskell.org for the first-time Haskell user


Martinsos:
Hi,
I am opening this issue as advised by @tomjaguarpaw , following up on the discussion I started on Haskell Foundation Slack.

I have been using Haskell for years but few days ago I wanted to check what is the current recommended way to install Haskell, hoping I might clean up my mess that somehow works but I can't even remember how I set it up (I am on ArchLinux, using Stack, didn't use ghcup so far).

In this process I actually ended up having a hard time figuring out what is the current recommended approach, which is I am writing this.

## My journey

I went to haskell.org first, but didn't see any obvious link/button that would guide me to setting up Haskell on my machine. I did see the "Downloads" link in the top navbar, but it wasn't emphasized in any way, and I usually expect from "Downloads" to be a list of binaries, which wasn't what I was looking for, I was looking for a "Getting started" kind of thing that will take me through the process of setting up Haskell.

I scrolled through the whole webpage and saw link to Haskell wiki at the bottom, so I concluded that might have section "Beginners" or "Getting started" or smth and I clicked on that.

On haskell wiki I saw catchy orange "Download Haskell" button on haskell wiki. That button took me to haskell.org/platform, where I went through some technical noise (details about stack upgrade and libgmp) to at the end learn that I should install `ghcup` and `stack`.

I then went to haskell.org/ghcup to learn that ghcup also installs stack, so started wondering if I should install stack via ghcup or if I should do it separately as previous instructions said.

At that moment I concluded this is all getting too complicated and maybe I went the wrong way, so I went back to haskell.org because I remembered I saw "Downloads" there, and maybe I should have clicked on that. "Downloads" took me to instructions that sent me to haskell.org/ghcup again.

At this point I gave up for now because I wasn't sure if I should use ghcup to install Stack or not, and I was also afraid it might mess up my hls (does ghcup do the "wrapper" thing to pick correct hls based on project? I wasn't sure), and decided to come back to it once I do more research and have more time.

## Thoughts

haskell.org -> "Downloads" -> haskell.org/ghcup is not a bad journey, but it still feels somewhat weird, as "Downloads" is in no way emphasized, and when I end up at ghcup page it feels a bit like this is a third-party project due to different webpage design and style. Plus, top part of GHCup home page is pretty complex and feels more aimed at contributors than first time users.

[...]

PR #208, opened 2022-08-05
Start coding section

### LINKED ISSUE #207: Add a "Start Coding" section.
Hi. It isn't a secret that newcommers have problems setting up a Development environment. The reason is that there is no single place within haskell.org, `hls` documentation nor `ghcup` documentation where you can find a **minimal, clear and strightforward guide** (MCSG from now on). 

Just to be clear and avoid unproductive discussion. Let me explain what a MCSG is:

- Because is a **guide** it should guide you step by step from _"My computer only has OS software installed"_ to _"I am programming in Haskell with the full power of an IDE"_
- Because is **strightforward** it should not branch. _you don't want to use `vscode`?_ ok, let me point you to "other editors" link, but such a link shouldn't be part of the guide itself.
- Because is **clear** each step should not be larger than one click or one execution command. 
- Because is **minimal** it should not give more information than necessary. If I have to navegate across multiple markdown hearders to find out how to configure the editor, or how to install the building tool, then it isn't minimal.

Some consequences of having a MCSG is that you need to select sensible defaults hence being biased towards tooling. In the Haskell case, the alternative which has proved better consistency is `ghcup` + `ghc` + `cabal` + `hls` + `vscode`. This isn't an opinion. `hls` is biased towards `vscode` and `stack` does not integrate well with `hls` unless you configure it properly (using `system-ghc`). 

If we were about to choose other editors or `stack` as default options, then we would need to provide extra configuration instructions for them, hence It wouldn't be **minimal**. If we were about to let the user decide between `stack` and `cabal` or `vim`/`emacs`/`vscode` then It wouldn't be **straightforward**. Lastly, if we don't use `ghcup` it would be **clear** because downloading each tool manually and configuring each of them to work properly and not break on updates is difficult if you aren't an experienced user.

The benefits of having a MCSG are clear. Many developers (In my experience, much more than I would have expect) do not know what a language server is, do not know or care about unix philosophy, do not want to read documentation, do not actually read documentation if they can copy paste a command in the terminal. If these developers come to haskell.org and find different ways of installing haskell, not mention about editors, many different tools, etc... the most probable outcome is "I will just installed pycharm and be happy". Actually, this is what I found helping newcommers (objectively, not a huge sample: a few people over cardano's discord and five co-workers). 

In [Discourse](https://discourse.haskell.org/t/get-a-haskell-development-environment-in-three-steps/4822) I posted a link to a [github gist](https://gist.github.com/lsmor/bb632565cd96be9da589b6e91f80f9ba) having a three step guide which is commited to MCSG as defined here. Whether it should be on `hls` home page or haskell.org is debatable, that's why I am opening this issue. If you consider this belongs to `hls` just tell me. In my opinion, haskell.org should provide this information, because is very unlikely that newcommers now what a language server is, hence there is low probability of them searching for `haskell language server` instead of just `haskell`.  

The guide ifself is only the first markdown header. There are two more sections including 

- "other editor" pointing to  `hls` documentation
-  Instructions for `stack` integration, because it is still a widely used tool. 

by the way, `stack` instructions are a little bit verbose because there is a surprisingly high amount of `stack` user who don't know how `stack` works. Every now and then there is a question in SO asking "[why](https://stackoverflow.com/questions/70267593/on-a-stacks-package-file-what-is-the-difference-between-library-and-executable/70272743#70272743) [stack](https://stackoverflow.com/questions/69374974/transformers-0-5-6-2-from-stack-configuration-does-not-match-0-3-0-2/69379846#69379846) [doesn't](https://stackoverflow.com/questions/62762252/installing-aeson-package-using-stack/62792161#62792161) [build](https://stackoverflow.com/questions/61673967/stack-build-error-while-building-package-aeson-cabal-exitfailure-1/61751248#61751248) [my project](https://stackoverflow.com/questions/72850322/how-does-the-workflow-with-haskell-stack-work/72855353#72855353)"

lsmor:
It closes #207 

relates to #189 and #136

At the moment it favors instalation via GHCup, vscode as an editor and cabal as a building tool.I think these are sensible defaults. I have put the link at the top row just because I found it the easiest way to do it, but let me know where would be the best position.


tomjaguarpaw:
Thanks for this! It will be a while before I get the chance to look at this, but I didn't want to leave this contribution without a response. Perhaps other committee members will get a chance to look before me.

lsmor:
Thanks for the reply. I'll be out-of-internet this month, so I am not in rush for the review :)

site/templates/nav.html
```diff
@@ -12,6 +12,7 @@
             <li $if(isCommunity)$     class="active" $endif$ ><a href="/community/">Community</a></li>
             <li $if(isDocumentation)$ class="active" $endif$ ><a href="/documentation/">Documentation</a></li>
             <li $if(isDonations)$     class="active" $endif$ ><a href="/donations/">Donate</a></li>
+            <li $if(isSetUp)$         class="active" $endif$ ><a href="/start-coding/">Start Coding now</a></li>
```
soupi, line 15: This should be the first link in the list.

lsmor: Done

TikhonJelvis (approved):
Thanks for the PR!

The content covered here *partially* overlaps with the downloads page, but that isn't necessarily a problem. Seems that having separate "getting started" and "downloads" pages could improve the overall design of the site by giving each page a clearer identity. "Getting Started" would be a "how-to"-style document for beginners while "Downloads" would be reference documentation for getting tools/GHC versions/etc.

I haven't thought this through in detail though; this is just a quick thought upon reviewing this PR.

lsmor:
@TikhonJelvis  Great to see this moving on! I see you've approved the changes but didn't merge yet. Do we need other mantainer's check? Do I need to change something?

Best,
Luis

TikhonJelvis:
Yeah, we need at least one more committee member to look it over. I mentioned this PR on our internal Slack channel so hopefully it'll get some attention soon :)

No changes needed from you, at least based on my review!

soupi (commented):
Consider linking to the documentation page as well so that the user can pick a learning resource

site/start-coding.markdown
```diff
@@ -0,0 +1,26 @@
+---
+title: Start Coding now
+page: start-coding
+isSetUp: true
+---
+
+## Install a development environment in three steps
+
+A Haskell development environment consist in:
```
tomjaguarpaw, line 9:
```suggestion
A Haskell development environment consists of:
```

site/start-coding.markdown
```diff
@@ -0,0 +1,26 @@
...
+
+- a compiler (`ghc`)
+- a language server (`hls`)
+- a building tool (`cabal` and/or `stack`)
```
tomjaguarpaw, line 13:
```suggestion
- a build tool (`cabal` and/or `stack`)
```

tomjaguarpaw (changes requested):

site/start-coding.markdown
```diff
@@ -0,0 +1,26 @@
+---
+title: Start Coding now
```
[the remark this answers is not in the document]
lsmor: ups! Start Coding Now looks better to me

site/start-coding.markdown
```diff
@@ -0,0 +1,26 @@
...
+- a building tool (`cabal` and/or `stack`)
+- an editor compatible with the language server protocol.
+
+It is recommended to use `ghcup` to manage your haskell toolchain and `vscode` as your editor.
+
+- **Step 1**: Install [GHCup](https://www.haskell.org/ghcup/). You'll be prompted to install some tools: Install `hls` and optionally `stack`. (`stack` users should [configure it properly](https://docs.haskellstack.org/en/stable/Stack_and_VS_Code/#workaround-1)).
+- **Step 2**: Install [vscode](https://code.visualstudio.com/). For integration with other editors see `hls` [documentation](https://haskell-language-server.readthedocs.io/en/latest/configuration.html#configuring-your-editor)
+- **Step 3**: Open `vscode` and install the [haskell extension](https://marketplace.visualstudio.com/items?itemName=haskell.haskell) either using the extension panel or pressing CTRL+P and `ext install haskell.haskell`.
+
+## Try without installing Haskell
```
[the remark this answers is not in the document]
lsmor: Oh!! I've been looking for this for a while. Thanks for pointing out

lsmor:
> Consider linking to the documentation page as well so that the user can pick a learning resource

I have addressed the other changes but I don't know what you mean be this one.

soupi:
@lsmor a beginner needs to do two things: 1. Set up a dev env, which this guide walks them through, 2. Pick up a learning resource, which this guide does not cover. I suggest linking to the documentation page at the end of the article so that the user can do (2).

lsmor:
@soupi Added a new section pointing to documentation page. On another topic, now that `stack` has updated to v2.9.1 it is very likely that _step 1_ of this guide gets [outdated](https://discourse.haskell.org/t/ann-stack-2-9-1/5078/5?u=lsmor) (because `ghcup` will configure `stack` for you), should this be merge when `ghcup` updates, or should we warn about it now?

site/start-coding.markdown
```diff
@@ -0,0 +1,34 @@
...
+
+- Try it in the [Haskell Playground](https://play-haskell.tomsmeding.com/)
+
+- Use a complete visual studio code environment in Gitpod.
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/gitpod-io/template-haskell)
+
+## Learn Haskell
+
+Now you are ready to code. Pick up a learning resource from the [documentation page](/documentation/) and start to learn Haskell. Also, you may want to familiarize yourself with the build tool of your choice and the library ecosystem. Check out sections [Manuals and guides](/documentation/#manuals-and-guides) and [Library Documentation](/documentation/#library-documentation) for this purpose.
```
soupi, line 34: Looks good. The only reservation I have is that some of the guides linked there are a bit outdated and mention cabal sandboxes. I think we have more modern guides that could be used ([1](https://www.haskell.org/ghcup/steps/), [2](https://gilmi.me/blog/post/2021/08/14/hs-core-tools)) but that's not necessarily in the scope of this PR I think.

hasufell:
> @lsmor a beginner needs to do two things: 1. Set up a dev env, which this guide walks them through, 2. Pick up a learning resource, which this guide does not cover. I suggest linking to the documentation page at the end of the article so that the user can do (2).

The documentation page is a mess and will leave a beginner confused with too many options that are not structured.

lsmor:
> The documentation page is a mess and will leave a beginner confused with too many options that are not structured.

There seems to be contradictory opinions here. I agree with documentation page being a mess. The problem is that as far as I can tell there is no "official" learning resource, and most FOSS resources are outdated. Should the "start coding section" promote one?, Should just reference to documentation page? should include links to ecosystem's tools like hoogle, hackage, etc... but no book at all?

I think there should be a section including a _how to start a new project_ including links to `cabal` and `stack` user guides, and probably some snippet. Also, there should be documentation about `hackage`, `stackage` and `hoogle` at least, since these are fairly standards within the community. 

The problem with the above, is that there isn't a book to go to. Probably LYAH is the "default" but it is outdated afaik. 

I propose:
- Add a _start new project_ section as defined above
- Add a _learn haskell_ section pointing to documentation (not the optimal solution, but the best if we want to avoid having two different pages with learning resources.)

Notice that _start new project_ section would contain links to `cabal`, `stack`, `hackage`, etc... (which documentation page also contains) Since using the build tool, and package repositories is a _must_ for newcommers, I don't think link duplication matters a lot here.

hasufell:
> Should the "start coding section" promote one?

IMO yes. GHCup homepage does exactly that.

lsmor:
> I think fixing the problems caused by the documentation doesn't have to be part of the scope of this PR.

I agree. It would be better if the documentation page gets a refresh and the start coding page just links to that. Of course thats a different PR and discussion.

hasufell:
> I think fixing the problems caused by the documentation doesn't have to be part of the scope of this PR.

My suggestion is to *not* link to the documentation page, unless it's fixed. This information will do newcomers no good.

lsmor:
> My suggestion is to _not_ link to the documentation page, unless it's fixed. This information will do newcomers no good.

Up to you guys. It does not matter to me as long as the three steps to set up the dev-env are clear, and that's already there. I am happy to push another commit removing it or modifying it but I think we should move this forward in one direction.

lsmor:
> I'm not in the comittee, so my opinion ultimately doesn't matter.

sad to hear that. So. who is responsable for this repo? Should I drop out the link to docs or not??

hasufell:
> > I'm not in the comittee, so my opinion ultimately doesn't matter.
> 
> sad to hear that. So. who is responsable for this repo? Should I drop out the link to docs or not??

https://www.haskell.org/haskell-org-committee/

So far, @tomjaguarpaw is the most active.

tomjaguarpaw:
> So far, @tomjaguarpaw is the most active.

Fair warning though, I have a fairly large stack of things to do so I don't think I'll be able to get to this one quickly.

lsmor:
I am closing this in favor of #214 which looks more promising.

PR #211, opened 2022-09-12
mercury is not a bank

rebeccaskinner:
Minor change to Mercury's copy on the testimonials section to comply with some marketing regulations on the use of the term "bank"


TikhonJelvis (approved):
Makes sense to me.

rebeccaskinner (commented):
I'm adding a comment of self-approval here.

PR #213, opened 2022-09-23
Add haskell.mooc.fi

stephen-smith:
I wasn't able to test this locally, buildAndWatch ended with an error `[ERROR] _cache/a98f5fd1dec8ada591e76ae4112817f9 for Hakyll.Core.Resource.Provider.MetadataCache/community.markdown/body: Store.set: invalid argument (invalid byte sequence)`, but I'm fairly sure it is correct, and this resource has been mentioned several times in the subreddit.


TikhonJelvis (approved):
Looks good to me.

The site built correctly for me in a Nix shell. I am not sure what caused the error that you encountered.

tomjaguarpaw (approved):

PR #216, opened 2022-10-10
fix indirect link (core libraries committee)

dschrempf:
Tiny PR. Fixes an indirect link to the Wiki and then to GitHub.


tomjaguarpaw:
Thanks! We'll need to wait for one more approval.

TikhonJelvis (approved):

PR #217, opened 2022-10-10
nit: punctuation

dschrempf:
Tiny PR.

To be honest, I do not understand the sentence (quote):
```
For Nix the package manager on the distribution NixOS and other OS:
```

Does this mean that this advice is applicable to the Nix package manager on all operating systems? If so, I suggest using:
```
For the Nix package manager:
```
or probably better
```
When using the Nix package manager please note:
```

(But then, only NixOS is referenced in the first point).

So maybe we want to fix that right away?

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -77,7 +77,9 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 <div id="collapse-nixos" class="collapse">
 <!-- This installation method is owned by @maralorn -->
 
-*For Nix the package manager on the distribution NixOS and other OS:*
+*[Nix] is the package manager for NixOS and can also be used on other Linux distros and macOS.*
+
+[Nix]: https://nixos.org
 
 * **Using ghcup on NixOS does not work.** Use the official packages instead.
 * You can get started by installing
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -77,7 +77,7 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 <div id="collapse-nixos" class="collapse">
 <!-- This installation method is owned by @maralorn -->
 
-*for Nix the package manager on the distribution NixOS and other OS*
+*For Nix the package manager on the distribution NixOS and other OS:*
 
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
This refers to https://www.haskell.org/downloads/ under "Show Nix/NixOS instructions"

I understand "for Nix the package manager on the distribution NixOS and other OS" to mean "for Nix (the package manager rather than the OS) on either the NixOS distribution, or any other distribution that Nix (the package manager) supports".

I agree it would be good to reword in a clearer way. @maralorn is the owner of this installation method. Perhaps he can chime in.

maralorn:
My intention was to inform unknowing readers that Nix the package manager is used on NixOS but also on other OS and that the documentation applies generally to both cases. With the obvious and explicit exception of the first bullet point.
This is general information which does not necessarily belong here, because people could also figuring it out be gooling "nix" and reading for 10 seconds, but I felt like it might help people reading about Nix for the first time. Especially if they somehow heard that something called "Nix" can be used to build some-weird-haskell-project-they-want-to-use and now they wonder what they need for that and if they need to switch to NixOS first or something like that …

I have no strong opinion about this and am open for suggestions. We could simply drop the introductory statement. We could make it longer to say

    Nix is the package manager of the distribution NixOS, but also available on other OS:

Feels clumsy, but is maybe less confusing.

site/downloads.markdown
```diff
@@ -77,7 +77,7 @@ Alternatively, many operating systems provide GHC, cabal and Stack through their
 <div id="collapse-nixos" class="collapse">
 <!-- This installation method is owned by @maralorn -->
 
-*for Nix the package manager on the distribution NixOS and other OS*
+*For Nix the package manager on the distribution NixOS and other OS:*
```
TikhonJelvis, line 80:
```suggestion
*[Nix] is the package manager for NixOS and can also be used on other Linux distros and macOS.*

[Nix]: https://nixos.org
```

@maralorn I like your suggestion for making the context a bit more explicit. If we're doing that, seems like it's worth linking to the Nix website and mentioning macOS explicitly.

TikhonJelvis (approved):
Works for me :)

We'll need another person to approve the change

PR #214, opened 2022-10-01
Adding a Getting Started page

Martinsos:
As a result of the [discussion I started on HF Slack](https://haskell-foundation.slack.com/archives/C01NJC2920Y/p1661439862151419), I created this PR to improve UX of haskell.org for the new Haskellers, that want to get started with Haskell.

The problem I am trying to solve is described in details in the discussion I linked above, so I won't repeat it all here but will try to put it shortly:
I believe that currently haskell.org is relatively unfriendly for the newcomers, mostly due to the unclear steps for getting started + somewhat confusing UX.
I believe we can fix this easily by copying what many other languages / webpages do, which is providing a "Getting started" page that takes beginners through setting up their dev environment, doing first steps in Haskell, pointing them to the communities they should join, and finally pointing them to the learning resources to continue with. While doing all this, it is important that path is clear and there are as little choices as possible.

I did the following main changes:
1. Added new /getting-started page that gives quickly gives a rundown of how to set up dev environment, run your first haskell program, join the community, and continue learning.
2. Added "Getting Started" button on the home page, that takes you to /getting-started page.
3. Added a "callout" to the top of /downloads page to ensure newcomers take a look at /getting-started page, since /downloads was before serving this purpose and I believe there are still quite few links on the internet leading to it.

Some notes:
1. I on purpose used only term `cabal` instead of `cabal-install` (CLI) or `Cabal` (lib) or `cabal format` (.cabal), because I believe it is too confusing to explain these differences to somebody who is taking a very first look at Haskell. In GHCup the term used is just `cabal`, and once they install it, they will call it with `cabal`, so for the start I think it is best to use just term `cabal` and later they will learn the difference.
2. I am not a native English speaker, but I did pull the /getting-started through Grammarly. I am open to any suggestions that improve wording!
3. I did my best regarding design while adding a Get Started button, but I am no Bootstrap expert so I would appreciate a review of my work there, especially regarding responsivity and how it works on smaller screens. I did on purpose add some more space between sections, to ensure that there is less stuff [above the fold](https://en.wikipedia.org/wiki/Above_the_fold), which should put more focus on the "Getting started" button.

@tomjaguarpaw @soupi 

EDIT (Oct 14th): updated photos to reflect new changes (fixes based on feedback from others).

image

image

image

Issue #212, opened 2022-09-23
Moving towards single installer


tomjaguarpaw:
Currently the [Downloads page](https://www.haskell.org/downloads/) has the following instructions:

1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
2. To install Stack, follow the [Stack installation guide](https://docs.haskellstack.org/en/stable/install_and_upgrade/)

It has been remarked elsewhere that this situation is undesirable. It would be much simpler and more friendly to have a big "Get Haskell" button. I'd like to move forwards towards that goal.

The first step is to identify what is the difference between the GHCup and Stack installers. Here is my current best understanding:

----

### Stack installer

* Can install Stack and GHC
* By default installs Stack (I'm not sure if it installs GHC upfront or if Stack installs GHC only as required by the user's packages)
* Doesn't install Cabal or HLS
* GUI installer on Windows, CLI installer otherwise.
* Installs GHC from official upstream bindists

### GHCup installer

* Can install GHCup, GHC, Cabal, Stack and HLS
* By default installs GHCup, GHC and Cabal
* CLI/TUI installer
* Installs GHC, Cabal and HLS from upstream bindists
* Does it install Stack from upstream bindists? If not, which bindist is it using?

----

There are three requirements to "merge" the functionality of these installers (but I think only a subset is worth doing) into a universal installer:

1. Ensure that the installer installs Stack from upstream bindists otherwise Stack maintainers might get bug reports from a bindist they don't support (perhaps GHCup already does this, I'm not sure)
2. Ensure that the installer installs GHC, Cabal and Stack by default
3. Ensure that the installer has CLI *and* GUI functionality on Windows

Personally I don't see much value in supporting 3.  I think we can achieve a satisfactory universal installer by making sure GHCup achieves 1 and 2. Perhaps it already satisfies 1. Satisfying 2 should be technically easy.

I would be interested in technical feedback on this idea, particularly from @hasufell. If it looks like GHCup can achieve 1 and 2 technically then with a bit of work I think we can achieve a consensus that that is the way forward.

----

### References

* <https://discourse.haskell.org/t/ghcup-button-on-www-haskell-org/4765>

hasufell:
> Ensure that the installer installs Stack from upstream bindists otherwise Stack maintainers might get bug reports from a bindist they don't support (perhaps GHCup already does this, I'm not sure)

[...]

PR #220, opened 2022-10-15
Recommend installing Stack via GHCup

tomjaguarpaw:
In my opinion, the time is right for www.haskell.org to recommend GHCup as the single install tool, for the following reasons:

* It is desirable for users to have a single way to install all of the Haskell toolchain. It is confusing to recommend they use two different installers.
* GHCup has proven to be a reliable installation tool for the Haskell toolchain.
* GHCup has been able to install Stack for the last 15 months
* GHCup [can now manage GHC versions for Stack](https://gitlab.haskell.org/haskell/ghcup-hs/-/merge_requests/285)
* GHCup installs both cabal-install and Stack by default, so it is not "biased" towards one or the other
* The current GHCup and Stack [have been collaborating](https://github.com/haskell-infra/www.haskell.org/issues/212#issuecomment-1272561149) to make sure the two tools play well together

----

This unlocks many follow ups, such as adding simple "getting started" or "start coding" sections, for example:

* https://github.com/haskell-infra/www.haskell.org/pull/214
* https://github.com/haskell-infra/www.haskell.org/pull/208

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -17,8 +17,7 @@ This page describes the installation of the Haskell toolchain, which consists of
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
 
-1. Install GHC, cabal-install and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
-2. To install Stack, follow the [Stack installation guide](https://docs.haskellstack.org/en/stable/install_and_upgrade/)
+* Install GHC, cabal-install, Stack and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
 
 * * *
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Procedurally, this will require a majority vote (4 out of 7) of the Haskell.org committee. I recommend that the committee votes in favour.

idabzo (approved):
LGTM, first timers needs non-optional guidelines and clear instructions.

romefeller (approved):

jaspervdj (approved):

TikhonJelvis (approved):
Yay, progress :)

PR #221, opened 2022-10-19
Emphasise ghcup, demphasise "toolchain"

tomjaguarpaw:
On this page we want it to be as easy as possible for people to see how to install Haskell. This PR switches to a single "call to action": Use GHCup.

It deemphasises the "Haskell toolchain", which is all very interesting, but shouldn't be the most prominent thing.

(There are lots of other, bigger, improvements we can make, of course, but there's so much low-hanging fruit now that I think we should just make a bunch of incremental improvements and we'll get to a good place.)

screenshot0

### DIFF
--- a/site/downloads.markdown
+++ b/site/downloads.markdown
@@ -17,7 +17,7 @@ This page describes the installation of the Haskell toolchain, which consists of
 
 *for Linux, macOS, FreeBSD, Windows or WSL2*
 
-* Install GHC, cabal-install, Stack and haskell-language-server via [GHCup](https://www.haskell.org/ghcup/)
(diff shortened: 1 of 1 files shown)


jaspervdj (approved):

ulysses4ever:
This is totally a step in the right direction. But the resulting design still feels like the main option is de-emphasized because of how much space it's given. It'd be worthwhile to add some text or stylistic elements to it, I think.

Compare to what rust does:
Screenshot_20221019-124802

Could we perhaps have a similar colored area to emphasize the main option? (Instead of green wet could have haskell.org-purple, I guess)

tomjaguarpaw:
Yeah, I think we should either have a big, obvious "Get Haskell" button or we should just use the GHCup `curl | sh` line on this page.

PR #223, opened 2022-11-05
Add a big friendly button: "Get Haskell"

tomjaguarpaw:
The button redirects to the downloads page for now, but we can update that if we add a getting started page and think that that's a better link target.

## Before

screenshot0

## After

screenshot1

Credit to @Martinsos


TikhonJelvis (approved):
Big and friendly all right :)

PR #225, opened 2022-11-07
Add alex's website

Ericson2314:
Putting my archivist hat on, I 

1. Download the darcs repo
2. Added uncomitted files
3. Converted it to git
4. Added files skipped in the conversion (? I don't know much about darcs)
5. Used `git subtree add` to put in `site/alex`
6. Added @andreasabel's revisions from the copy in the alex repo

I don't know if we want those binary files in *this* repo's history, but I think it would be good store them before blowing away that directory on the server. (One concrete reason to keep is that old bootstraps are very delicate, for example.)


andreasabel (approved):
I skimmed this PR.  Thanks!
If something could be wished for, that would be a `README.md` under `site/alex` explaining the content.  If I understand correctly, only `index.html` is hooked up, the rest is for historic purposes.
I wonder whether `html-OLD` can be deleted, it seems to be auto-generated from the old docbook xml and is maybe anyway duplicated in one of the archives.

tomjaguarpaw:
This is great, thanks! I took the liberty of pushing a commit that adds the `alex/` path to the builder, so these things actually appear on the website, which I assume is the intention.

Ericson2314:
Thanks @tomjaguarpaw! I forgot how to do that.

@andreasabel I wanted to put everything in the history, but we can just delete most stuff now. Tell me what it anything is worth keeping, since you last modified the page, and I'll do it.

tomjaguarpaw:
@Ericson2314 It looks like this PR is ready to be merged now, is that right?

Ericson2314:
@tomjaguarpaw I think so! We can improve the website more thereafter.

tomjaguarpaw (approved):

tomjaguarpaw:
Great, we'll have to wait for another committee member to approve.

Dear Haskell.org committee members, this MR move the website's `alex/` subdirectory into our repository. Currently it is just some files on the server, with no version control or oversight. Managing them from the main website repo sounds like a much better idea.

PR #226, opened 2022-11-07
Add happy's website

Ericson2314:
Similar to #225, I 

1. Download the darcs repo
2. Added uncomitted files
3. Converted it to git
4. Added files skipped in the conversion (? I don't know much about darcs)
5. Used `git subtree add` to put in `site/happy`.

I don't know if we want those binary files in *this* repo's history, but I think it would be good store them before blowing away that directory on the server. (One concrete reason to keep is that old bootstraps are very delicate, for example.)


tomjaguarpaw:
Thanks! Like https://github.com/haskell-infra/www.haskell.org/pull/225#issuecomment-1308378813 I assume we want to add these to the builder, so that they appear on the website.

tomjaguarpaw:
@Ericson2314 It looks like this PR is ready to be merged now, is that right?

Ericson2314:
I think so!

tomjaguarpaw (approved):

tomjaguarpaw:
Great, we'll have to wait for another committee member to approve.

Dear Haskell.org committee members, this MR move the website's `happy/` subdirectory into our repository. Currently it is just some files on the server, with no version control or oversight. Managing them from the main website repo sounds like a much better idea.

TikhonJelvis (approved):
Nice! The HTML/CSS style seems pretty old-fashioned... which makes sense given the commits for it go back 17 years :P

Thanks for getting the contents in this repo; it'll definitely make it easier to improve the contents down the line.

jaspervdj (approved):

PR #228, opened 2022-11-27
Delete Alex and happy websites, they should redirect to docs

Ericson2314:
(Plan from https://github.com/haskell/alex/issues/215, for example.)

The old websites in addition to being in this repo's history are in `old-www` branches in both projects' repos. That should be enough redundancy.

That actual adding of the redirects is done out of band, so this does the other steps.


Ericson2314:
The docs are https://haskell-alex.readthedocs.io/ and https://haskell-happy.readthedocs.io/, respectively.

tomjaguarpaw:
Can we please do this via an `http-equiv` in a `.html` file in the repo? I would rather have this repo eventually be the authoritative source for the entirety of `www.haskell.org` rather than some opaque configuration file on the server.

tomjaguarpaw:
Specifically, I think it should be something like

```.html
 <html xmlns="http://www.w3.org/1999/xhtml">    
  <head>      
    <title>Alex</title>      
    <meta http-equiv="refresh" content="0;URL='https://haskell-alex.readthedocs.io/'" />    
  </head>    
  <body> 
  </body>  
</html>
```

but having said that, perhaps we want the whole subsite to be redirected, in which case this PR is already doing the appropriate thing. I would welcome a brief discussion of the pros an dcons of each approach.

TikhonJelvis:
Having more structure in this repo sounds like a great idea for future maintenance and improvements. Does using http-equiv have any specific downsides compared to a server redirect?

jaspervdj:
I'm fine with having the redirect done on the server side iff this server configuration also lives here in of our repos.  I don't believe this is the case, so I like Tom's suggestion to use a `http-equiv` redirect (even though this isn't ideal for non-html user-agents, but I guess that's negligible market share).

Ericson2314:
I would like it to work the same way as Haddock which is a whole subsite redirect.

I agree out-of-band configuration is bad, and I would be happy to help fix that if we were to, say, use the nginx NixOS module to make that stuff declarative, but I think that should be done in a follow-up PR.

tomjaguarpaw:
OK, fair enough. That's convincing for me.

tomjaguarpaw (approved):

gbaz:
Redirects added to the server.

Since there's some confusion: we do use nginix nix modules to make the redirects as declarative as possible. And those modules are under source control. However, the nix configuration for the servers isn't in a public repo, and it might be possible to do so, but it would be tricky, as some components are fine to be public, and others are not.

tomjaguarpaw:
We need another approval from a committee member.

TikhonJelvis (approved):

Ericson2314:
Thanks!

tomjaguarpaw:
Hmm, that doesn't sound right. It should only be trying to remove things in its own directory, that were put there by a previous website PR.

tomjaguarpaw:
It may be related to this:

https://github.com/haskell-infra/www.haskell.org/blob/15b4fdbc1b4b27f71f97f8faf7d4de59087b45f9/deploy.sh#L19-L20

tomjaguarpaw:
But if it is that it means that the `sftp` `chmod` has failed. I don't know why that would be the case. Probably someone should go and poke around on the server but I probably won't have time to until the weekend.

tomjaguarpaw:
And for reference, this is the failed deploy: https://github.com/haskell-infra/www.haskell.org/actions/runs/3650401404/jobs/6166342473#step:6:98

Ericson2314:
Perhaps because the directories were made manually first, they have different perms?

tomjaguarpaw:
They weren't made manually. They were made automatically by the deploy of your previous MR that added them to the repo. The old `alex` and `happy` subsites are on a completely different directory on the server.

tomjaguarpaw:
I've had a look on the server and I can't see anything in the website contents directory, specifically `/www-data/haskell.org/www/`, that could explain why some files failed to be removed. Maybe there's something wrong with the `sftp` script that means files can't be removed at all. I noticed a similar problem on [a previous action](https://github.com/haskell-infra/www.haskell.org/actions/runs/3650399195).

tomjaguarpaw:
Oh, hang on, it's probably because we had two concurrent jobs trying to deploy at the same time. In which case the next deploy will bring things to a consistent state and we don't need to bother fixing anything unless we want to avoid this concurrent deploy problem, which is probably more trouble than it's worth at the moment.

TikhonJelvis:
That sounds plausible. We can also trigger the deploy workflow manually—which [I just did][1]—so we can see whether it gets fixed right away.

[1]: https://github.com/haskell-infra/www.haskell.org/actions/runs/3670232452

TikhonJelvis:
Okay, that job ran successfully, so we should be fine except for concurrent jobs. Glad that you figured it out :)

tomjaguarpaw:
That deploy was clean, so I suspect concurrent deploys was indeed the cause.

tomjaguarpaw:
Concurrent replies :D

Issue #229, opened 2022-12-05
Link for Haskell Art is dead


advalt:
on this page https://www.haskell.org/community/
this link https://lurk.org/groups/haskell-art/

TikhonJelvis:
Thanks! I can open a PR to remove the dead link.

Do you know if there's a new page for the same group? No worries either way—if there's a new link, we can always add that as a separate PR.

tomjaguarpaw:
Fixed by https://github.com/haskell-infra/www.haskell.org/pull/230

PR #230, opened 2022-12-05
Fixed link to Haskell Art mailing list

TikhonJelvis:
Looks like the current link leads to a 404. I found a URL to the mailing list page on we.lurk.org that works.

Addresses #229


tomjaguarpaw:
Thanks!

PR #236, opened 2023-01-21
Explicitly set locale to C.UTF-8 inside buildAndWatch script

TikhonJelvis:
This should fix the problem reported in #29

### DIFF
--- a/buildAndWatch
+++ b/buildAndWatch
@@ -12,6 +12,7 @@ function buildAndWatchWithNix() {
         exit 1
     fi
 
+    export LC_ALL=C.UTF-8       # fix locale error with Hakyll (see #29)
     nix-build -A builder && \
         ./result/bin/haskell-org-site clean && \
         ./result/bin/haskell-org-site build && \


Martinsos:
Yup, fixed!

tomjaguarpaw:
Do we understand why https://github.com/haskell-infra/www.haskell.org/pull/30 isn't enough? Do we want to *undo* #30 at the same time as merging this? I'm a bit uncomfortable setting environment variables all over the place, especially seemingly redundantly.

TikhonJelvis:
#30 only sets the environment variable during the Nix build and in a Nix shell. That used to be fine since `buildAndWatch` used to only *exist* inside the project's Nix shell, but at some point we rewrote the script so that it only runs *outside* a Nix shell. Instead, the new script uses Nix to build the executable but then runs it in a normal environment which may or may not have the right kind of locale settings.

A better solution would probably be rewriting the Haskell code so that it did not depend on the system's locale settings, but getting that right seems like a bit of a pain—the way locale stuff works in the standard library seems pretty fiddly, and it might even depend on Hakyll internals.

buildAndWatch
```diff
@@ -12,6 +12,7 @@ function buildAndWatchWithNix() {
         exit 1
     fi
 
+    export LC_ALL=C.UTF-8       # fix locale error with Hakyll (see #29)
```
rebeccaskinner, line 15: Rather than exporting this for the entire script we could run the individual commands with this set. For example:
`nix-build -A builder && LC_ALL=C.UTF8 ./result/bin/…`

that would prevent the change from leaking outside of the command that needs it.

tomjaguarpaw:
Personally I'd prefer to understand exactly which part of the pipeline needs the envvar, and set it only there, but approving, since I'm sure you know what you're doing.

tomjaguarpaw (approved):

rebeccaskinner (approved):
I agree with @tomjaguarpaw that we can probably just merge this for now and maybe revisit later see if we can remove the need for it.

TikhonJelvis:
Probably best to set the env var for any invocation of `./result/bin/haskell-org-site` for consistency. We can always change the approach if it becomes a problem in the future.

PR #242, opened 2023-01-29
Fixed landing page design a bit (bootstrap errors, spacing, alignment).

Martinsos:


site/index.html
```diff
@@ -5,22 +5,27 @@
 ---
 <div class="header">
    <div class=" container ">
-      <div class=" row ">
-         <div class=" span12 col-sm-12 hidden-xs"><br></div>
+      <div class="row">
+         <div class=" span12 col-sm-12 hidden-xs"><br><br></div>
+      </div>
+      <div class="row" style="display: flex; align-items: center">
```
Martinsos, line 11: Here I did 3 changes:

1. I added additional `<br>`. There was one before, now it is two. I added this `<br>` and a couple of others below in some places in order to put the "get started" button close to the center of the page, and push Videos section closer to the bottom of it. I did this guided by the typical advice for designing a landing page -> you want your CTA to be the most attractive thing to click that is visible [above the fold](https://www.optimizely.com/optimization-glossary/above-the-fold/). In our case that is "getting started" button, potentially also purple "try" section. Videos section is still partially visible above the fold, but is a bit more down and therefore looks more like smth additional to explore than one of the main things to look at.

2. I closed the div that is Bootstrap row here, instead of letting it also capture the stuff below. One Bootstrap row is always organized as 12 columns, but it was misused here, because somebody first added an element that spans all 12 columns, and then below one more element that spans 6, and then one more that spans 6!
The way Bootstrap handles such situation result in ok result, but it is semantically incorrect. So I closed this div here, so it contains only one element of 12 columns, and then opened a new one below (that is the last line). This element of 12 columns is anyway used to add vertical empty space in the whole width of the page, so it makes perfect sense that it is a standalone row.
This is merely a Bootstrap fix of the previous mistake somebody did.

3. In the new "row" I created, I used the opportunity to align Haskell logo with the code example to the right by their centers. Before they were aligned by their tops. That wasn't that much of a bit deal before when there was no "get started" button, but now with that button added, height of haskell logo vs code example + get-started button has bigger difference than before, which resulted in a bit weird look in my opinion.

site/index.html
```diff
@@ -5,22 +5,27 @@
...
          <div class=" span6 col-md-6">
             <div class="branding">
-               <br class="hidden-xs"><img src="/img/haskell-logo.svg" class="img-responsive">
+               <br class="hidden-xs">
+               <img src="/img/haskell-logo.svg" class="img-responsive">
```
Martinsos, line 15: Here I merely formatted the html a bit.

site/index.html
```diff
@@ -5,22 +5,27 @@
...
+                     <pre><span class='hs-definition'>primes</span> <span class='hs-keyglyph'>=</span> <span class='hs-varid'>filterPrime</span> <span class='hs-keyglyph'>[</span><span class='hs-num'>2</span><span class='hs-keyglyph'>..</span><span class='hs-keyglyph'>]</span>
+      <span class='hs-keyword'>where</span> <span class='hs-varid'>filterPrime</span> <span class='hs-layout'>(</span><span class='hs-varid'>p</span><span class='hs-conop'>:</span><span class='hs-varid'>xs</span><span class='hs-layout'>)</span> <span class='hs-keyglyph'>=</span>
+              <span class='hs-varid'>p</span> <span class='hs-conop'>:</span> <span class='hs-varid'>filterPrime</span> <span class='hs-keyglyph'>[</span><span class='hs-varid'>x</span> <span class='hs-keyglyph'>|</span> <span class='hs-varid'>x</span> <span class='hs-keyglyph'>&lt;-</span> <span class='hs-varid'>xs</span><span class='hs-layout'>,</span> <span class='hs-varid'>x</span> <span class='hs-varop'>`mod`</span> <span class='hs-varid'>p</span> <span class='hs-varop'>/=</span> <span class='hs-num'>0</span><span class='hs-keyglyph'>]</span></pre>
+                  </div>
```
Martinsos, line 28: Here I just added a bootstrap "row", in line 20, because it was missing, which is a mistake from before. Everything else is not a change, just indentation changed a bit so git thinks its new.

site/index.html
```diff
@@ -35,7 +40,13 @@ <h4 class="tag">Declarative, statically typed code.</h4>
       </div>
    </div>
 </div>
-<br><br class="hidden-xs hidden-sm"><br class="hidden-xs hidden-sm">
+
+<br>
+<br class="hidden-sm">
+<br class="hidden-sm">
+<br class="hidden-xs hidden-sm">
+<br class="hidden-xs hidden-sm">
```
Martinsos, line 48: Here I added some extra spacing, as explained above. So before it was 3 brs, now it is 5. Notice that just one happens on any screen size, other 2 happen only when screen is > sm, and other 2 happen only when screen > xs. So, smaller the screen, less spacing, which is what we want.

site/index.html
```diff
@@ -63,7 +74,13 @@ <h2>Try it!</h2>
       </div>
    </div>
 </div>
+
 <br>
+<br class="hidden-sm">
+<br class="hidden-sm">
+<br class="hidden-xs hidden-sm">
+<br class="hidden-xs hidden-sm">
+
```
Martinsos, line 83: Same as with spacings above.

tomjaguarpaw:
Personally I'm not particularly keen on this. For me it's a question of "if it ain't broke, don't fix it" and I don't see that there's anything particularly broken here. I'd be interested to know what others think.

TikhonJelvis (approved):
Looks like a reasonable set of improvements to me. The small visible tweaks are an improvement, and it seems better to match with bootstrap's semantic model if we can.

Martinsos:
> Personally I'm not particularly keen on this. For me it's a question of "if it ain't broke, don't fix it" and I don't see that there's anything particularly broken here. I'd be interested to know what others think.

Makes sense, although I would argue that in this instance it is broken, just hasn't manifested yet -> here I am referencing to incorrect usage of bootstrap rows.

tomjaguarpaw:
Yes, fixing the bootstrap thing makes more sense. I think it can afford to be separated into its own PR.

Martinsos:
Ok, I opened a new one with just bootstrap stuff here: https://github.com/haskell-infra/www.haskell.org/pull/243 .

I adjusted this PR to be the same as the bootstrap one + 1 commit extra that brings landing page improvements via spacing -> so the same thing as it was, but changes it does are clearly delineated in commits. If I could I would have made it to point to bootstrap PR but I can't because it is on my fork.

tomjaguarpaw:
Looks like some merge conflicts got into here after the force push.

Martinsos:
> Looks like some merge conflicts got into here after the force push.

Thanks, my mistake, should be ok now.

tomjaguarpaw:
Unfortunately this seems to introduce a problem, because on narrow screens the "Haskell logo" block is forced to stay next to the "primes" block.

screenshot0

The status quo is that they wrap nicely:

screenshot0

Martinsos:
@tomjaguarpaw auf that is an issue indeed, thanks for this and I should have checked that.

I fixed it now by removing the flexbox styling I did to align Haskell logo with the code on the right.

I am not adamant that this PR needs to be merged -> I thought adding some spacing might be useful, but if you all feel it doesn't help much, I am ok with closing it, and somebody else can address the design in a better, more knowledgeable way, I am not very adept in design.

PR #245, opened 2023-02-02
Use haskell.org IRC page

ffaf1:
haskell.org has a pretty good IRC page, lean, informative, with clear onboarding instructions. Let us use that instead of the one on the wiki.


TikhonJelvis (approved):
Makes sense to me.

tomjaguarpaw:
Oh yes, thanks!

PR #247, opened 2023-02-03
State that GHCup does not require root access

ffaf1:
When I try new software, I am always wary of installers requiring root access, as I don't want to turn my system into a [FrankenDebian](https://wiki.debian.org/DontBreakDebian). Specifying that GHCup is well behaved in this regard might be of importantce to a sizeable portion of users.

I would like this information to be present, but also to be as unobtrusive as possible (it is a "Get Started!" page, saddling it with too many things is counterproductive), so I welcome suggestions on where to put this.

### DIFF
--- a/site/get-started.markdown
+++ b/site/get-started.markdown
@@ -26,7 +26,7 @@ A complete Haskell development environment consists of the Haskell toolchain (co
 
 ### GHCup: universal installer
 
-[GHCup](https://www.haskell.org/ghcup/#) is a universal installer for Haskell that will install everything you need to program in Haskell, and will help you manage those installations (update, switch versions, ...). Follow the instructions on the [GHCup webpage](https://www.haskell.org/ghcup/#) to install GHCup. Then use it to install the Haskell toolchain, which consists of:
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
Can you suggest some wording for the Downloads or Get Started pages that would be suitable?

ffaf1:
For "Get Started": see patch! I see “Downloads” now has:

    Use GHCup to install GHC, cabal-install, Stack and haskell-language-server

I feel this probably is short enough not to warrant to anything else. It is not an explanatory paragraph, you are supposed to click on the link to get more informations.

tomjaguarpaw:
> For "Get Started": see patch!

Sorry, which patch?

ffaf1:
This shows as a pull request, right?

tomjaguarpaw:
Oh I'm sorry, I thought it was an issue! Yes, thanks, I see the suggested wording in the PR.

tomjaguarpaw:
I often have the same concerns as you regarding root access so I like the idea of calling this out. What do others think?

TikhonJelvis:
Hey, sorry, totally missed this PR earlier, but it sounds like a reasonable addition.

TikhonJelvis (approved):

Issue #252, opened 2023-02-03
Haskell "try it" not working on Android


andys8:
There is a prompt that lets you try Haskell on the landing page. Input doesn't seem to work on Android in mobile browsers. I tested on 2 devices (Samsung and OnePlus) with 2 different browsers (Kiwi and Chrome). On the other hand it *is* working on iOS (Chrome and Safari).

https://user-images.githubusercontent.com/13085980/216621119-2bc84307-c5ed-4d95-8a9a-6aeea22605ec.mp4

andys8:
The underlying implementation is https://github.com/chrisdone/jquery-console. The project isn't maintained and there are no newer versions available.

But luckily there are 2 open pull requests suggesting a solution to the Android issue.

* https://github.com/chrisdone/jquery-console/pull/71
* https://github.com/chrisdone/jquery-console/pull/68

So, a fix seems to be possible. The alternative would be to hide "try it" on Android.

tomjaguarpaw:
FWIW the "try it" app is very limited and there have been suggestions to replace it with a proper [Play Haskell](https://play-haskell.tomsmeding.com/).

tomjaguarpaw:
See https://github.com/haskell-infra/www.haskell.org/issues/185

tomjaguarpaw:
You can give it a go if you like!

PR #253, opened 2023-02-03
Fix try haskell on Android

andys8:
## Fix

Cherrypicked fix provided in https://github.com/chrisdone/jquery-console/pull/71.

Tested on desktop, iOS, and Android - with Android now working. There is also another [comment](https://github.com/chrisdone/jquery-console/pull/71#issuecomment-364102640) confirming the fix.

Fixes https://github.com/haskell-infra/www.haskell.org/issues/252

## Review

I would welcome another person making sure "try it" is still working on desktop as before (comparing with currently deployed https://haskell.org).

Hint for review: I needed to change the listening host to `0.0.0.0` in order to be able to access the local dev server via my mobile device in the same network.

```patch
diff --git a/builder/site.hs b/builder/site.hs
index c5869dd..a56877b 100644
--- a/builder/site.hs
+++ b/builder/site.hs
@@ -66,7 +66,7 @@ main = mkContext >>= \ctx -> hakyllWith configuration $ do
 
 
 configuration :: Configuration
-configuration = defaultConfiguration{ providerDirectory = "site" }
+configuration = defaultConfiguration{ providerDirectory = "site", previewHost = "0.0.0.0" }
 
 parseTestimonialCompiler :: Compiler (Item Testimonial)
 parseTestimonialCompiler = do
```

## Demo

https://user-images.githubusercontent.com/13085980/216684916-37d97c75-b324-4978-9beb-a2388bd3e8d7.mp4

### DIFF
--- a/site/js/jquery.console.js
+++ b/site/js/jquery.console.js
@@ -283,13 +283,26 @@
 	    }, 0);
 	});
 
+    //helper function to figure out new character added to input
+    difference = function(value1, value2) {
+        var output = [];
+        for (i = 0; i < value2.length; i++) {
+            if (value1[i] !== value2[i]) {
+                output.push(value2[i]);
+            }
+        }
+        return output.join("");
+    }
+
 	////////////////////////////////////////////////////////////////////////
 	// Handle key hit before translation
 	// For picking up control characters like up/left/down/right
 
 	typer.keydown(function(e){
+        typer.oldValue = typer.val();
+        oldValue = typer.val();
 	    cancelKeyPress = 0;
-	    var keyCode = e.keyCode;
+        var keyCode = e.which || e.code;
 	    // C-c: cancel the execution
 	    if(e.ctrlKey && keyCode == 67) {
 		cancelKeyPress = keyCode;
@@ -311,12 +324,7 @@
 		    return false;
 		}
 	    }
-	});
 
-	////////////////////////////////////////////////////////////////////////
-	// Handle key press
-	typer.keypress(function(e){
-	    var keyCode = e.keyCode || e.which;
 	    if (isIgnorableKey(e)) {
 		return false;
 	    }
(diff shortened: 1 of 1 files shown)


tomjaguarpaw:
For me, on desktop, it doesn't seem to allow entering space or numerals.

andys8:
Oh, man, then it's not as easy as applying the suggested fix. Will take a look :D

andys8:
I pushed a change where copied the [file from the fork](https://github.com/chrisdone/jquery-console/blob/85fab01f3fcbf77158993037d93068e0e9d88e86/jquery.console.js) over as is.

The result seems to be that it's working on Android and Desktop. Before I could confirm the same behavior you described.

Why didn't I do this in the first place? The fork has some formatting chances, so I thought I could only use the lines that fix logic. But also: The [file of the fork has 845 lines](https://github.com/chrisdone/jquery-console/blob/85fab01f3fcbf77158993037d93068e0e9d88e86/jquery.console.js) while the file in the [haskell-infra repository has only 750 lines](https://github.com/haskell-infra/www.haskell.org/blob/master/site/js/jquery.console.js). The difference is significant. I haven't looked into it and don't know if these differences are improvements, or code that was actively stripped away to run it on haskell.org.

tomjaguarpaw:
The latest version seems to be fine on desktop.

site/js/jquery.console.js
```diff
@@ -37,714 +37,809 @@
...
+            while (deleteCharAtPos()) {
+                updatePromptDisplay();
+            }
+        };
+
+        function clearCurrentPrompt() {
+            extern.promptText("");
+        };
+
+        function clearScreen() {
```
andys8, line 444: :information_source: This change is coming from this commit: https://github.com/chrisdone/jquery-console/commit/3d405bf1f1613e5c456159adfb6d1b13512b196a

site/js/jquery.console.js
```diff
@@ -37,714 +37,809 @@
...
+                }
+            } else { // Assume it's a DOM node or jQuery object.
+                inner.append(msg);
+            }
+            newPromptBox();
+        };
+
+        ////////////////////////////////////////////////////////////////////////
+        // Report some message into the console
+        function report(msg, className) {
```
andys8, line 588: :information_source: This change is coming from this commit: https://github.com/chrisdone/jquery-console/commit/c8812e4a2a3a4328775391bff06a77aa42753bd0

andys8:
@tomjaguarpaw Good to merge?

TikhonJelvis (approved):
Makes sense to me, but would be great to have @tomjaguarpaw give it a quick test too.

tomjaguarpaw:
When I tried it on desktop it was fine so let's give this a go. I don't really understand how it works, but if anyone spots any problems after merge we can always roll back.

andys8:
>  if anyone spots any problems after merge we can always roll back

Exactly. Thanks for merging

tomjaguarpaw:
Thanks for contributing!

Issue #249, opened 2023-02-03
buildAndWatch fails because of locale / invalid byte sequence


andys8:
## Issue

```shell
/nix/store/2md452mq5labdcxlsc5gfik97dhxabbr-haskell-org-0.1.0.0
Removing _site...
Removing _cache...
Removing _cache/tmp...
Initialising...
  Creating store...
  Creating provider...
  Running rules...
Checking for out-of-date items
Compiling
  updated 2015-11-10-Haskell-Org-Board-Resolution.pdf
  [ERROR] _cache/a98f5fd1dec8ada591e76ae4112817f9 for Hakyll.Core.Resource.Provider.MetadataCache/community.markdown/body: Store.set: invalid argument (invalid byte sequence)
```

## Related

* People pointing out the error is related to UTF8 encoding and hakyll:
https://github.com/jaspervdj/hakyll/issues/614
* Similar issue in nixpgks: https://github.com/NixOS/nixpkgs/pull/122117#discussion_r1058479008
* #29
* #236 
* https://github.com/haskell-infra/www.haskell.org/issues/246#issuecomment-1415836754

## System

I'm using OS: Manjaro 22.0.0 Sikaris, arch based.
Nix 2.8.1 installed as package manager, but not OS.

https://wiki.archlinux.org/title/locale

## Locale

```shell
$ locale -a
C
C.UTF-8
de_DE.utf8
en_DE.utf8
en_US.utf8
POSIX

$ locale
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_NUMERIC=de_DE.UTF-8
LC_TIME=de_DE.UTF-8
LC_COLLATE="en_US.UTF-8"
LC_MONETARY=de_DE.UTF-8
LC_MESSAGES="en_US.UTF-8"
LC_PAPER=de_DE.UTF-8
LC_NAME=de_DE.UTF-8
LC_ADDRESS=de_DE.UTF-8
LC_TELEPHONE=de_DE.UTF-8
LC_MEASUREMENT=de_DE.UTF-8
LC_IDENTIFICATION=de_DE.UTF-8
LC_ALL=
```

TikhonJelvis:
[...]

PR #255, opened 2023-02-03
Explicitly set local encoding to UTF-8 in site.hs

TikhonJelvis:
This change makes sure that the site executable *always* expects UTF-8, regardless of any locale environment variables. This should avoid problems like #29 and #249 in a more robust way than trying to set the right environment variables in `buildAndWatch` or `shell.nix`.

The diff ended up a bit noisy because of indentation changes; the key change was adding `Encoding.setLocaleEncoding Encoding.utf8` to the beginning of `main`.

I tested this by reproducing the Unicode error (using `nix-shell --pure`) and checking that the Haskell code change fixes the problem.

### DIFF
--- a/buildAndWatch
+++ b/buildAndWatch
@@ -12,7 +12,6 @@ function buildAndWatchWithNix() {
         exit 1
     fi
 
-    export LC_ALL=C.UTF-8       # fix locale error with Hakyll (see #29)
     nix-build -A builder && \
         ./result/bin/haskell-org-site clean && \
         ./result/bin/haskell-org-site build && \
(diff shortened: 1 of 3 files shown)


TikhonJelvis:
In particular, this should help us avoid problems where Nix [does not provide `en_US.UTF-8` by default][1] while distros like RHEL 7 [do not support `C.UTF-8`][2].

[1]: https://github.com/NixOS/nix/issues/4829
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1590680

tomjaguarpaw (approved):
Looks reasonable. These locale things have always confused me. I'm keeping my fingers crossed that this one does the trick!

rebeccaskinner (approved):
Thanks for digging into this; seems good to me.

andys8:
Can confirm this branch is working on my 2 linux systems:

## Manjaro (based on Arch)

```shell
```

## Mint (based on Ubuntu/Debian)

```shell
```

TikhonJelvis:
Perfect, thanks for testing, I'll merge this now.

PR #262, opened 2023-03-29
Add redirects for old urls

### LINKED ISSUE #261: Redirects for old "A History Of Haskell" urls?
(This is an old issue, originally [opened here](https://github.com/haskell-infra/hl/issues/217); thought I'd see if it's still something fixable. Also, it doesn't looks like redirects are managed in this repo, but if they are somehow I'd be happy to take a stab at a PR.)

I am reading A History Of Haskell: Being Lazy With Class. There are several instances of URLs that point to 404s, all on haskell.org, that perhaps could be redirects.

- haskell.org/humor (should be: https://wiki.haskell.org/Humor)
- https://www.haskell.org/hat -> https://wiki.haskell.org/Hat
- http://haskell.org/development/views.html -> ???

mjhoy:
These redirects are for links in the paper "A History Of Haskell: Being Lazy With Class."

The `createRedirects` function produces html with meta tag redirects ([source](https://jaspervdj.be/hakyll/reference/src/Hakyll.Web.Redirect.html)). I think this should properly handle the following URLs:

- https://www.haskell.org/humor -> https://wiki.haskell.org/Humor
- https://www.haskell.org/hat -> https://wiki.haskell.org/Hat
- https://www.haskell.org/development/views.html -> http://web.archive.org/web/20040107202217/http://haskell.org/development/views.html

Fixes #261.

### DIFF
--- a/builder/site.hs
+++ b/builder/site.hs
@@ -56,6 +56,12 @@ main = do
                      ctx
         defCompiler indexCtx
 
+    let redirects = [ ("humor/index.html", "https://wiki.haskell.org/Humor")
+                    , ("hat/index.html", "https://wiki.haskell.org/Hat")
+                    , ("development/views.html", "http://web.archive.org/web/20040107202217/http://haskell.org/development/views.html")
+                    ]
+    version "redirects" $ createRedirects redirects
+
     match ("**/*.markdown" .||. "*.markdown") $ do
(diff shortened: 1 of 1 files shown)


TikhonJelvis (approved):
Nice, yeah, looks good to me.

Issue #263, opened 2023-04-30
LInk to University Virginia course rediects to an inappropriate site


preveen-stack:
https://github.com/haskell-infra/www.haskell.org/blob/cef565184391c1f3e765d6858cf1e53ab354c78b/site/documentation.markdown?plain=1#L44

tomjaguarpaw:
Thanks for the report!

PR #264, opened 2023-04-30
Remove dead link

tomjaguarpaw:
Closes https://github.com/haskell-infra/www.haskell.org/issues/263


Kleidukos (approved):

TikhonJelvis:
PR looks good but cachix is broken; I can take a look at what's going on later today if nobody else can fix it before then.

tomjaguarpaw:
I think we need to fix https://github.com/haskell-infra/www.haskell.org/issues/266 before we can CI properly.

TikhonJelvis (approved):

PR #265, opened 2023-04-30
New version of cachix-action

tomjaguarpaw:

PR #267, opened 2023-04-30
Update install-nix-action and cachix-action

### LINKED ISSUE #266: Cachix action is failing
See, for example https://github.com/haskell-infra/www.haskell.org/actions/runs/4845420159/jobs/8634334648#step:4:13

I thought that https://github.com/haskell-infra/www.haskell.org/pull/265 might fix it, but it does not: https://github.com/haskell-infra/www.haskell.org/actions/runs/4845428718/jobs/8634343440?pr=265

@TikhonJelvis or @rebeccaskinner, as the Nix experts, would you mind taking a look?

TikhonJelvis:
Updating install-nix-action should fix #266 (see [cachix @144][1]). I also noticed that `deploy.yml` was not using cachix-action at all, so I updated that as well.

Since we're updating the workflows anyhow, I figure we can also bump `cachix-action` while we're at it—same change as #265 but applied to both workflow files.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,7 +20,13 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Install Nix
-        uses: cachix/install-nix-action@v12
+        uses: cachix/install-nix-action@v20
+
+      - name: Cachix
+        uses: cachix/cachix-action@v12
+        with:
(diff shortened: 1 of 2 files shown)


tomjaguarpaw (approved):

tomjaguarpaw:
Great, thank you for fixing that quickly!

PR #268, opened 2023-05-04
Remove Google Analytics tracking

Kleidukos:


tomjaguarpaw:
Great, thanks. Since this might count as a "controversial change" we formally need at least 4 approvals (including the author), and we now have them. It might be prudent to wait a bit in case any other committee member wants to give their input.

tomjaguarpaw (approved):

TikhonJelvis (approved):
Sounds good to me.

rebeccaskinner (approved):
LGTM

PR #271, opened 2023-06-27
Fix styling of links

alt-romes:
In trying to improve the appearence of the Haskell main webpage a little bit, I've
1. Removed the underline of links that appear
    * inside buttons
    * in the navbar
2. Changed the dotted underline to a solid underline

Screenshots:

Old:

Screenshot 2023-06-27 at 13 59 32

Screenshot 2023-06-27 at 13 59 42

Screenshot 2023-06-27 at 13 59 49

New:

Screenshot 2023-06-27 at 11 07 50

Screenshot 2023-06-27 at 11 07 40

Screenshot 2023-06-27 at 11 07 34

I suppose the dotted vs solid underline decision is more contentious, so I've added it as a separate commit we can also just drop.

Thanks!

### DIFF
--- a/site/css/hl.css
+++ b/site/css/hl.css
@@ -275,7 +275,7 @@ textarea {
 
 a {
   color: #9E358F;
-  text-decoration: underline dotted; }
+  text-decoration: underline ; }
   a:hover, a:focus {
     color: #65225b;
     text-decoration: underline; }
--- a/site/css/hl.css
+++ b/site/css/hl.css
@@ -2032,6 +2032,9 @@ select[multiple].input-lg,
     -webkit-box-shadow: none;
     box-shadow: none; }
 
+a.btn {
+  /* No text-decoration for <a> tags in buttons */
+  text-decoration: none; }
 a.btn.disabled,
 fieldset[disabled] a.btn {
   pointer-events: none; }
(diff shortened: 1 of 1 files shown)


TikhonJelvis (approved):
Thanks!

Based on the screenshots I'd be fine with changing the dotted underline to a normal underline. Curious what other folks think.

tomjaguarpaw:
I think we should also drop the underline entirely on the sponsor names and the video titles. They don't seems to serve any purpose to me and I think solid underline looks much worse there.

TikhonJelvis:
We generally need something more than just color to indicate links, both for accessibility reasons and because we use purple all over the place. The video titles seem sufficiently clear based on context so we could drop the underline there, but I'm not sure about the sponsor names.

tomjaguarpaw:
Ah yes, I was thinking the whole sponsor rectangle was a link, but it's only the logo and name. Well, I think the full underline is far too obtrusive in the sponsor name link, but I'm not a graphic designer, so weight my opinion accordingly.

tomjaguarpaw (approved):

tomjaguarpaw:
Looks good to me. @TikhonJelvis I'll let you carry this over the line.

TikhonJelvis:
Sweet. Thanks for the improvements @alt-romes!

PR #275, opened 2023-07-17
Hide tracking pixel

emtes:
**Issue**
This recent [PR](https://github.com/haskell-infra/www.haskell.org/pull/274) added a Scarf tracking pixel. While the link is indeed to a 1px by 1px image, it's currently taking a lot of space at the bottom of the page because it's a flex item.

cc: @TikhonJelvis (stakeholder; added the pixel)

**Summary of changes:**
- remove pixel from flow of page with absolute positioning
- use existing `invisible` class to add `visibility:hidden`

**Testing**
- ensured the pixel is still being loaded by looking at Network tab
- clicked around the site to make sure nothing else started looking funky
---
Demos
**CURRENT**
Screenshot from 2023-07-17 14-06-59

**AFTER**
Screenshot from 2023-07-17 14-16-51

### DIFF
--- a/site/templates/default.html
+++ b/site/templates/default.html
@@ -61,6 +61,6 @@
      <script src="/js/jquery.console.js"></script>
      <script src="/js/tryhaskell.js"></script>
      <script src="/js/tryhaskell.pages.js"></script>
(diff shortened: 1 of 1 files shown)


aaronallen8455 (approved):

TikhonJelvis (approved):
Sweet, thanks for spotting and fixing the problem quickly!

PR #278, opened 2023-08-03
Serve fonts from www.haskell.org

### LINKED ISSUE #203: Remove Google Analytics and Google Fonts?
[Recent rulings in the EU declared Google Analytics unlawful.](https://noyb.eu/en/update-further-eu-dpa-orders-stop-google-analytics)

I think keeping it there is a huge liability, not to mention an invasion of privacy. Hackage already removed it in https://github.com/haskell/hackage-server/pull/1097 (effective on next deploy).

I propose to remove it from haskell.org too.

Google fonts is also problematic, and using it instead of serving the fonts from haskell.org does not bring any benefit since browsers partition their caches by domain nowadays.

emtes:
**Summary:** This PR removes the network call to the Google Fonts API in favor of font files hosted with our other server assets. Closes #203 

**Testing:** Pull local and deployed versions of the site side by side to compare

**Risks:**
- Loading fonts from wherever this site is hosted might be slower
    * **Mitigation:** Use `font-display: swap;` to tell the browser that it shouldn't wait too long for the fonts, and use them when they do arrive
- To copy the `fonts` directory to the built site, I pattern matched some existing code. It worked, and at a glance seems fine, but I don't _really_ understand `site.hs`.
    * **Mitigation:** Review the Haskell change carefully

**Notes:** 
- The site was using "Source Sans Pro", a sans serif font no longer listed on Google Fonts. Best I can tell, "Source Sans 3" is the same thing. This change is likely responsible for a tiny pixel shift in the new build, but everything still looks good.
- The Google Fonts API call had links for other glyphs like Vietnamese and Greek. The site only uses Latin characters, so only those were added to the repo.

| BEFORE | AFTER |
|--------|--------|
| image | image |
| image | image |
| image | image |


site/templates/default.html
```diff
@@ -8,8 +8,8 @@
       <meta content="haskell,functional,pure,programming,lazy" name="keywords">
       <meta content="The Haskell purely functional programming language home page." name="description">
       <link href="/img/favicon.ico" rel="shortcut icon">
-      <link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro|Raleway:700,900|Ubuntu+Mono:400" type="text/css" rel="stylesheet">
-      <link href="/css/hl.css" type="text/css" rel="stylesheet">
+      <link rel="stylesheet" href="/css/fonts.css">
```
emtes, line 11: `hl.css` is unruly at 4.7k+ lines of unformatted CSS, so I think for now separate stylesheets for the fonts is okay. In the future, maybe we can break out `hl.css` using something like Sass modules that can be compiled into a single CSS file at build time?

emtes:
@tomjaguarpaw for review?

site/templates/default.html
Kleidukos: Yes that would be nice to have an assets pipeline that deals with this.

Kleidukos (approved):
LGTM

TikhonJelvis:
Sweet, thanks for the fix!

PR #281, opened 2023-09-05
update "Effective Haskell" entry

egmaleta:
"Effective Haskell" is not in beta stage anymore, since august 1st


TikhonJelvis (approved):
Nice, thanks for the update!

tomjaguarpaw (approved):

PR #283, opened 2023-09-27
Fix small text a11y issues

emtes:
Mainly, adding the `alt` attribute where needed.
`alt` with no value means the image is for decoration only, like the video thumbnails.


TikhonJelvis (approved):
Looks great, thanks!

Kleidukos (approved):

PR #284, opened 2023-09-28
"Features" section improvements

enmcarv:
This PR proposes:
- removing unused code from `home.js`
- changing the copy of the features section's collapse buttons to say “Show more”
- change collapse buttons to have a different copy when toggled (“Show less”)

What do you all think of the copy changes? I think this is a slight improvement from “Click to expand” since the styling of the link already indicates the user can click there. Alternatively, we could have “Click to expand” and “Click to hide”.

An outstanding issue is that each feature should probably indicate that it can be clicked to toggle collapse outside the link button. Once resolved, there might not be a need for a separate button. A designer's opinion would be great :sweat_smile: 

Screenshot from 2023-09-28 16-28-39

[Screencast from 2023-09-28 16-28-52.webm](https://github.com/haskell-infra/www.haskell.org/assets/54204155/00ddeb65-4c78-470b-bf79-e87ed981182c)


TikhonJelvis (approved):
Looks great, thanks! I like the way "show more" looks and it seems just as clear as the previous wording.

PR #285, opened 2023-10-14
CIS 194 domain has changed

master-of-none:
The website domain has been changed.
Fixed it to point to the right course of 2013, since 2023 course has no materials.


TikhonJelvis (approved):
Thanks for the fix!

Kleidukos (approved):

PR #286, opened 2023-11-29
add RFP for site redesign

rebeccaskinner:
This change adds the contents of the haskell.org redesign RFP to the site. It's not linked from the main site. We can link to the page directly, giving us an easy way to share the RFP across multiple sites.

Although the technical change here is pretty safe, I'll hold off on merging this until we get a couple of reviews from other people on the committee.


aaronallen8455 (approved):

site/rfp/haskell-org-redesign-rfp.markdown
ryantrinkle, line 1: Might be good to add a deadline.

TikhonJelvis (approved):

Issue #288, opened 2024-02-29
Update videos on homepage


kosmikus:
I'll be honest: My primary motivation is to make "The Haskell Unfolder" (https://www.youtube.com/playlist?list=PLD8gywOEY4HaG5VSrKVnHxCptlJv2GAn7) more visible.

However, I generally think that the videos currently on the homepage have been there for a long time, and if we want to make Haskell look like an actively used language with a vibrant community, we should aim to choose recent videos even if the old ones are without doubt good. It just makes a better impression to a visitor of the page if the videos are not all a decade old.

Here are a few suggestions for good videos and playlists that I could quickly come up with. My perspective is limited, I'm almost certainly forgetting many good options. I've deliberately included playlists, and videos targeted at different levels / backgrounds. My suggestion would be to generally update more often, or possibly even have a larger selection and randomly pick a few on every visit (the SPJ video is currently there already; I think it's good, and I didn't find a more recent and equally relevant SPJ video, so I'd keep it):

Functional Programming & Haskell - Computerphile (John Hughes)
https://www.youtube.com/watch?v=LnX3B9oaKzw

Escape from the Ivory Tower / The Haskell Journey (Simon Peyton Jones)
https://www.youtube.com/watch?v=re96UgMk6GQ

Past and Present of Haskell - Interview with Simon Peyton Jones
https://www.youtube.com/watch?v=4RuLzL_q0zs

The Haskell Unfolder (Andres Löh and Edsko de Vries)
https://www.youtube.com/playlist?list=PLD8gywOEY4HaG5VSrKVnHxCptlJv2GAn7

Functional Programming in Haskell (Graham Hutton)
https://www.youtube.com/playlist?list=PLF1Z-APd9zK7usPMx3LGMZEHrECUGodd3

Advanced Functional Programming in Haskell (Graham Hutton)
https://www.youtube.com/playlist?list=PLF1Z-APd9zK5uFc8FKr_di9bfsYv8-lbc

Laziness in Haskell (Alexis King)
https://www.youtube.com/playlist?list=PLyzwHTVJlRc8620PjqbM0x435-6-Gi1Gu

Delimited Continuations Demystified (Alexis King)
https://www.youtube.com/watch?v=DRFsodbxHQo&list=PLOvRW_utVPVmzDGGOJ2amgVBK168Vemke&index=3

What is a Monad? - Computerphile (Graham Hutton)
https://www.youtube.com/watch?v=t1e8gqXLbsU

[...]

PR #290, opened 2024-03-01
Update videos.

kosmikus:
This is resulting from the discussion at #288 and implements the compromise suggested by @tomjaguarpaw in that thread.


tomjaguarpaw (approved):
Thanks! We'll need to wait for one more reviewer before merge.

Tritlo:
Looks good to me! I like the selection also, much more familiar than the previous one.

PR #308, opened 2024-07-20
Remove some dormant/defunct conferences

frasertweedale:
With two Aussie confs on this list, and having attended and presented at Curry On 2018, I take no pleasure at all in submitting this PR :(


Kleidukos (approved):

Kleidukos:
Condolences.

Issue #310, opened 2024-08-20
LYAH link outdated on https://www.haskell.org/documentation/


tomjaguarpaw:
cc @Bodigrim

PR #311, opened 2024-08-20
Update LYAH link to a working one

tomjaguarpaw:
Similar to https://github.com/haskell-infra/www.haskell.org/pull/298


tomjaguarpaw:
This will resolve https://github.com/haskell-infra/www.haskell.org/issues/310

aviaviavi (approved):

PR #313, opened 2024-10-01
show how to run `ghci` without installing permanently

fricklerhandwerk:
This change reduces the number of steps required to run `ghci` to two:
- install Nix (need to press `y` and type `sudo` password)
- run `gchi` in a temporary environment

We can somewhat safely assume that most experienced Haskell users are also using Nix (evidence: ZuriHac 2023 show of hands got >90%), so having Nix around to work with Haskell is not wrong.
Wiring up full Haskell development environments with Nix is also a lot easier than without, and even the minimal case of having a shell with the `ghc` and `cabal-install` packages pinned to specific versions makes life much, much simpler.

Disclaimer: I'm a Nix maintainer and also took responsibility for Nix beginner documentation. Which in turn means I could help with improving the onboarding experience for new Haskell users by e.g. publishing a tutorial.

### DIFF
--- a/site/get-started.markdown
+++ b/site/get-started.markdown
@@ -6,16 +6,49 @@ isGettingStarted: true
 
 # Get started
 
-Welcome, new Haskeller! Read on to quickly set up your Haskell development environment, execute your first lines of code, and get directions for further learning.
+Welcome, new Haskeller! Read on to run your first lines of code, set up your Haskell development environment, and get directions for further learning.
 
 ## Content
+
+  - [Running your first lines of code](#running-your-first-lines-of-code)
   - [Set up a Haskell development environment](#set-up-a-haskell-development-environment)
     - [GHCup: universal installer](#ghcup-universal-installer)
     - [Editor](#editor)
-  - [Running your first lines of code](#running-your-first-lines-of-code)
   - [Writing your first Haskell program](#writing-your-first-haskell-program)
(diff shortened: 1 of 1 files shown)


Kleidukos:
I am not in favour. ZuriHac is a particularly skewed population of Haskell aficionados. This population is not the one that this page is targeting. I would rather orient people to the [Playground](https://play.haskell.org) for an online experience.

fricklerhandwerk:
Well alright, but then there need to be very different instructions that integrate with the playground. Right now there seems to be nothing in between "ephemeral hacking in the browser" and "do these 50 steps until you can compile a module".

Kleidukos:
Let us not talk about 50 steps, because this solution does force Windows users to use WSL2 and leaves them on the side of the road regarding Windows-native toolchains, which ghcup supports.

fricklerhandwerk:
Okay. Just trying to figure out if it makes sense to improve this contribution, and if so, how exactly, or whether to just close the PR. It seems not obvious to me how one can point to the playground without rewriting the entire page and then still leave users hanging because the playground doesn't have any instructions on it.

hasufell:
> install Nix

Why would we expose new users to the worst fiddling experience since linux package managers exist?

Bodigrim:
> and then still leave beginners hanging because the playground doesn't have any instructions on it.

I've raised https://github.com/haskell/play-haskell/issues/52 to discuss this.

PR #292, opened 2024-05-01
Add Logo option UwU

HEIGE-PCloud:
## Overview

Adds a query param to switch the Haskell logo to https://github.com/SAWARATSUKI/ServiceLogos:

- on: `?uwu=true`
- off: `?uwu=false`
Setting will persist until you turn it off.

## Screen

327063873-4bb93066-6606-4fab-9c5c-c54c94836630

## Credits 

Credit to @SAWARATSUKI in https://github.com/SAWARATSUKI/ServiceLogos

## Context

- https://react.dev/?uwu=true
- https://hono.dev/?kawaii=true
- https://twitter.com/saltyAom/status/1785339588661461365

PR #320, opened 2024-11-22
Revert alternative logo

rhendric:
Six months ago:

‘It would most certainly be temporary...’


tomjaguarpaw (approved):
Fine by me

MangoIV:
But why would we revert it?

tomjaguarpaw:
Exactly because [part of the justification for adding it](https://github.com/haskell-infra/www.haskell.org/pull/292#issuecomment-2090806936) was that it would be temporary. It would set a bad precedent to keep something that was explicitly justified on the grounds of being temporary. If people _do_ want to keep it, that's fine, but I think it should go through another proposal process.

(In retrospect I think we ought to have set a time limit on it, after which it would have been "pre-approved" for reversion, absent another vote to keep it.)

tomjaguarpaw:
> I think it's a fun Easter egg so I would say keep it.

I'm inclined to agree actually. It seems fun and I don't think it's doing any harm. But it is up to the people who want to keep it permanently to make such a proposal and advocate for it, otherwise we're going to lose trust in proposals whose proponents claim "this is only temporary".

kutyel:
I think it does not hurt anyone if we keep it and also it shows that the Haskell community still has a nice sense of humour 😉

PR #322, opened 2024-12-05
Add alternative logo (pride)

rhendric:
Haskell logo in the colors of the Pride Progress flag

If we're going to have easter eggs that promote subcultures, I'd want us to start here.

Suggestions for improvement are very welcome.

This is a derivative both of the existing Haskell logo (licensed under the BSD 3-Clause, AFAIK) and the Progress Pride flag (licensed for non-commercial use under Creative Commons BY-NC-SA 4.0, with [extra text waiving the requirement for attribution](https://progress.gay/pages/terms-of-use) ([archive.org snapshot](https://web.archive.org/web/20250506001126/https://progress.gay/pages/terms-of-use))).

BSD 3-Clause is a permissive license, and per the [license compatibility checker](https://interoperable-europe.ec.europa.eu/licence/compatibility-check/BSD-3-Clause/CC-BY-NC-SA-4.0), it is acceptable to relicense a derivative of a BSD 3-Clause work under BY-NC-SA. It is unclear to me if the text that waives the requirement for attribution for the Progress Pride flag is considered part of the license, and thus subject to the ShareAlike clause of BY-NC-SA and required to be part of the licensing of derivative works, but I am happy to abide by it as if it were.

Therefore, I hereby offer this logo under the [Creative Commons BY-NC-SA 4.0](https://creativecommons.org/licenses/by-nc-sa/4.0/deed.en) license, and similarly to the Progress Pride flag's terms of use linked above, I waive the requirement to credit me when using or creating derived works from this logo.

PR #323, opened 2024-12-05
Add alternative logo (retro)

rhendric:
Animated Haskell logo in the style of an early 90s handheld gaming device

This art is my own work (except for the Haskell wordmark which is derived from the SVG Haskell logo), and I explicitly license it under the BSD 3-Clause license used by the rest of the site.


tomjaguarpaw:
Thanks for the submission, @rhendric. After deliberation the committee has decided not to accept new logo submissions at this time. Please see the following for some more details:

https://github.com/haskell-infra/www.haskell.org/pull/322#issuecomment-2903872820

PR #331, opened 2025-03-25
make last example less whitespace sensitive

haydenflinner:
Fixes a little frustration I hit while working through the site's built-in tutorial. Eventually I clicked the spoiler-answer to figure out what I was missing :-)

whitespace sensitive haskell tutorial

### DIFF
--- a/site/js/tryhaskell.pages.js
+++ b/site/js/tryhaskell.pages.js
@@ -518,7 +518,7 @@ tryhaskell.pages.list =
             "<code>let abc@(a,b,c) = (10,20,30) in (abc,a,b,c)</code>"
         },
          trigger:function(result){
-             return result.expr.match(/^[ ]*let[ ]*\(_,\(?a:_\)?\)[ ]*=[ ]*\(10,\"abc\"\)[ ]*in[ ]*a[ ]*$/) &&
+            return result.expr.match(/^[ ]*let[ ]*\(_,[ ]*\(?a:_\)?\)[ ]*=[ ]*\(10,[ ]*\"abc\"\)[ ]*in[ ]*a[ ]*$/) &&
                  result.type == "Char";
          }},
         {guide:function(result){


tomjaguarpaw:
Could you spell out explicitly what this does? I don't understand yet.

TikhonJelvis:
If I'm understanding the code correctly, the interactive "Try Haskell" tutorial uses regular expressions to check if you entered the "right" expression before going on to the next step. The regular expression for the last step was too restrictive, so a valid expression like `let (_, (a:_)) = ...` would not be accepted, while `let (_,(a:_)) = ...` would.

I'd probably go as far as adding optional spaces around the `:`, because I tend to write `(a : _)` instead of `(a:_)`.

By the way, @haydenflinner, thanks for submitting this fix! I always appreciate people who quick fixes as soon as they run into problems.

site/js/tryhaskell.pages.js
```diff
@@ -518,7 +518,7 @@ tryhaskell.pages.list =
             "<code>let abc@(a,b,c) = (10,20,30) in (abc,a,b,c)</code>"
         },
          trigger:function(result){
-             return result.expr.match(/^[ ]*let[ ]*\(_,\(?a:_\)?\)[ ]*=[ ]*\(10,\"abc\"\)[ ]*in[ ]*a[ ]*$/) &&
+            return result.expr.match(/^[ ]*let[ ]*\(_,[ ]*\(?a:_\)?\)[ ]*=[ ]*\(10,[ ]*\"abc\"\)[ ]*in[ ]*a[ ]*$/) &&
```
TikhonJelvis, line 521:
```suggestion
            return result.expr.match(/^[ ]*let[ ]*\(_,[ ]*\(?a[ ]*:[ ]*_\)?\)[ ]*=[ ]*\(10,[ ]*\"abc\"\)[ ]*in[ ]*a[ ]*$/) &&
```

I tend to write `(a : _)` instead of `(a:_)` myself, so let's catch that too

tomjaguarpaw:
Oh I see. So the screenshot is of _corrected_ behavior, not the old behavior?

tbidne:
I believe the screenshot is of the current behavior, though it is hard to see the actual problem itself.

The user in the screenshot is attempting to write an expression that pattern matches on the 'a'.

The first try does not typecheck, due to the tuple vs. list syntax. The second and third attempt _do_ typecheck -- and correctly solve the exercise -- but they do not match the actual expectation, which is apparently whitespace-sensitive. The fourth attempt matches the expectation, hence the congratulatory "Perfetto..." text. But the second and third should have counted as well.

You cannot really tell this from the screenshot unless you happen to know that the congratulatory text only showed up after the last attempt. I just tried it though; that's exactly what happens.

tomjaguarpaw:
Wow, that is bizarre behavior! Why not judge the input to be a success if it returns the correct value, with whatever syntax it was obtained? Anyway, I guess we should merge this and include @TikhonJelvis's suggestion too.

tbidne:
I'd guess because the syntax is part of the exercise: the user demonstrating they know how to use pattern matching to extract the `'a'` from `(10, "abc")`. Otherwise they could just write `'a'`.

Though you're right that the current expectation is quite strict (another example: the variable must be named `a`). It's probably hard to be more flexible without massively complicating it.

It might be nice to include some kind of notification for various failures, since right now there is nothing. The only indication that it's not correct is that the prompt text does not change, which will probably lead to what the OP did here: reveal the answer, debug why it went wrong. For instance, I could imagine:

- Incorrect result (but no type error): "Sorry, that is not the right answer."
- Correct result, regex failure: "The result is correct, but that's not quite what we were looking for."

haydenflinner:
> If I'm understanding the code correctly, the interactive "Try Haskell" tutorial uses regular expressions to check if you entered the "right" expression before going on to the next step. The regular expression for the last step was too restrictive, so a valid expression like `let (_, (a:_)) = ...` would not be accepted, while `let (_,(a:_)) = ...` would.

Exactly it! Though the good news for a beginner is, I was confident enough after [this video](https://www.youtube.com/watch?v=RcVA8Nj6HEo) and the Haskell tutorial up until this step that I was convinced I must be right, it must be bugged 😂 I'm taking it as a good indicator for the language, even if the change is in pragmatic JS 😄 .


> I'd probably go as far as adding optional spaces around the `:`, because I tend to write `(a : _)` instead of `(a:_)`.
> By the way, @haydenflinner, thanks for submitting this fix! I always appreciate people who quick fixes as soon as they run into problems.
Thank you, and likewise!

tomjaguarpaw (approved):
Well, I approve since this is clearly an incremental improvement.

tomjaguarpaw:
We'll need to wait for another committee member to approve.

Kleidukos (approved):

tomjaguarpaw:
Thanks for this @haydenflinner!

PR #337, opened 2025-05-09
Focus on Haskell benefits in main (sub)headline

malteneuss:
Follow-up from discussions on https://discourse.haskell.org/t/emphasize-why-haskell-on-haskell-org-landing-page/12036 

Motivation: Increase general Haskell adoption by **focusing on the business/developer benefits in headlines and descriptions, and to reduce technical jargon** such that decision makers and developers from other programming languages understand the points easily, and to nudge them to try out Haskell.

image

Specific arguments for the chosen wording:

- "**Next-generation**": Emphasize that Haskell has (and has had) powerful features that don't exist yet in mainstream languages and may only trickle down much later. I'm thinking of higher-kinded types, GADTs, Linear Types, and hopefully dependent types.
- "**maintainable**": Seems to be a common experience that this is the most important benefit of using Haskell, e.g. see [Simon Pyton Jones](https://discourse.haskell.org/t/emphasize-why-haskell-on-haskell-org-landing-page/12036/12?u=malteneuss)
- "**reliable**": A one-word benefit summary of "If it compiles, it works" experience many Haskellers share.
- "**efficient**": I would mention it, because environmental concerns and cloud costs are important nowadays and from my experience many production-grade Haskell libraries like persistent, Yesod, conduit, Servant etc. are efficient by default compared to e.g. Java, Python.
- "**solid foundations**": Benefit summary of all the mathematical concepts behind Haskell, phrased in a way that everyone can understand. I would mentioning "purely functional", which doesn't mean anything to someone that hasn't used it, and sounds scary/complicated at first glance.

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -14,7 +14,7 @@ isHome: true
                <br class="hidden-xs alternative-logo-hidden">
                <img src="/img/haskell-logo.svg" alt="Haskell" class="img-responsive alternative-logo-hidden">
                <img src="/img/haskell-uwu.png" alt="Haskell UwU" class="alternative-logo alternative-logo-visible">
-               <h4 class="summary">Write next-generation maintainable, reliable and efficient software on solid foundations</h4>
+               <h4 class="summary">Enjoy long-term maintainable software you can rely on</h4>
             </div>
          </div>
          <div class=" span6 col-md-6">
--- a/site/index.html
+++ b/site/index.html
@@ -14,7 +14,7 @@ isHome: true
                <br class="hidden-xs alternative-logo-hidden">
                <img src="/img/haskell-logo.svg" alt="Haskell" class="img-responsive alternative-logo-hidden">
                <img src="/img/haskell-uwu.png" alt="Haskell UwU" class="alternative-logo alternative-logo-visible">
-               <h4 class="summary">An advanced, purely functional programming language</h4>
+               <h4 class="summary">Write next-generation maintainable, reliable and efficient software on solid foundations</h4>
             </div>
          </div>
          <div class=" span6 col-md-6">


hasufell:
I actually like the old summary better.

The main page also has the "Features" section. I'd envision a similar section called "Why Haskell?", which is focused less on specific language features, but more on the "development experience".

Both are important.

malteneuss:
Then we agree to disagree (on the summary) and need a governance process to come to a decision.

To me the current headline is equivalent to as if https://www.rust-lang.org/ used "Rust: An advanced systems-programming language with a borrow-checker", which doesn't convey much to people not already familiar with the language.

malteneuss:
Good point, the fun part was mentioned a lot in the discussions. Just adding a few variations into the mix:

"Next-generation maintainable, delightful and reliable software<br/>on solid foundations"
image

"Next-generation delightful, maintainable and scalable software<br/>on solid foundations"
image

aviaviavi:
I love this PR! Even if we already have some disagreement about exactly _what_ this tagline should say, improving our elevator pitch here is a very good use of collective efforts. 

I personally find this to be a big improvement. Much more value oriented, which is important. Agree that what we have here is just a tad long, perhaps because of the "next-generation" piece which is more vague than the rest. 

A few ideas: 
- Write maintainable, reliable, and efficient software on solid foundations. (just dropping the next-generation piece)
- Write code that's built to last: maintainable, reliable, efficient.

Overall I'd vote we definitely try to get the best variant of this merged, and bet it will make a meaningful difference to the % of visitors who continue on to try it out.

malteneuss:
I like the "Write maintainable, reliable, and efficient software on solid foundations." as well.
image

Bodigrim:
* I'd put more emphasis on maintainability, like "Write perpetually maintainable, reliable and efficient software"
* "Write" and "software" are so nineties. How about "Build perpertually maintainable and reliable products"?
* Could use a positive / playful angle indeed. "Build perpertually maintainable products and have fun"? "Discover joy of perpetually maintainable products"?

ninioArtillero:
For this to move forward I propose a general guideline for anyone joining:

1. Read the forum thread through.
2. Keep referencing previous ideas to give them momentum and introduced a new element only if necessary. I think we have plenty on the table to get this done.
3. Let's just have at most 3 elements-adjectives as a hard constraint on sub-headline variants.

> * "Write" and "software" are so nineties. How about "Build perpertually maintainable and reliable products"?
> 
> * Could use a positive / playful angle indeed. "Build perpertually maintainable products and have fun"? "Discover joy of perpetually maintainable products"?

I like the idea of moving from "software" as it is also kind of redundant, but "products" feels too capitalist-industrial for my taste. May be such noun could be implicit.

> The main page also has the "Features" section. I'd envision a similar section called "Why Haskell?", which is focused less on specific language features, but more on the **"development experience"**.

The sub-headline should not mention any specific features (which Haskell has a whole lot), but rather convey the experience, power and community such features afford. Why Haskell? Is a question whose answer has many dimensions, all worth addressing upfront on the landing page. The sub-headline should rather convey a feeling.

> I would perhaps suggest: "A fun language for building software that lasts"

The human element is very important and this gets a good balance with the enterprise value. But I think it lacks the Haskell personality. Which I'll relate to elegance and being at the forefront of PL. For me using "fun" here is very revolutionary and cool, but seems difficult to mix-in without coming across as too informal or lax.

It is impossible to get everything in, considering space constraints: This should be succinct as well! I'ld dare suggest this variant:

"A pioneering language for fearless development and enduring reliability."

Or breaking my own proposed rule and stubbornly try to get all the goodies:

"A pioneering language for fearless development, enduring reliability, and joyful expression."

malteneuss:
> I'd put more emphasis on maintainability, like "Write perpetually maintainable, reliable and efficient software"

Thinking about people with English as a 2nd language, i would fall back to simpler vocabulary than "perpetually" (i like the messaging though) or "pioneering" if possible; although i can't think of a good adverb for maintainable. 

> * "Write" and "software" are so nineties. How about "Build perpertually maintainable and reliable products"?

Indeed "Build" or "Enjoy" (to emphasize the fun aspect a bit more formally) as a verb are good replacements for "Write". I also would like to avoid generic "software", but can't think of a better word. Haskell is not (yet) famous for "products".

How about?
"**Enjoy maintainable, reliable and efficient software**" (maybe a bit dull, but concise and conveys many Haskell experiences), 
or a bit more abstract 
"**Enjoy maintainable software you can depend on**" (throw out the efficient subtext)

Bodigrim:
> Thinking about people with English as a 2nd language, i would fall back to simpler vocabulary than "perpetually" (i like the messaging though) or "pioneering" if possible; although i can't think of a good adverb for maintainable, maybe "lifelong"?

As a non-native speaker, I'd say that both "perpetual" and "pioneering" are pretty basic vocabulary. Maybe "eternally maintainable" then, "lifelong" does not sound quite right to me. "Long-term maintainable" is also good.

> I also would like to avoid generic "software", but can't think of a better word. Haskell is not (yet) famous for "products".

One more option is "projects", although I'm not sure it's an improvement.

> "Enjoy (lifelong) maintainable, reliable and efficient software"

No programming language ever advertised itself an unmaintainable, unreliable and inefficient; I think we really need additional qualifiers (such as "lifelong") to stand out. 

> "Enjoy (long-term) maintainable software you can depend on" (throw out the efficient subtext)

This is my favorite so far. +1 to throwing "efficient" out, I don't think Haskell can claim any particular achievements for efficiency. I'd prefer "you can rely on" or "rely upon" over "depend on".

hasufell:
I think the headline should be short and convey the most basic information that makes Haskell a different language. 

And the current one does exactly that: it's a lazy purely functional programming language. 

That's why Haskell came into existence in the first place: to explore those properties.

There's no reason to wash it down with sales speak.

noughtmare:
I don't think it is fair to dismiss the proposed alternatives in this thread as "sales speak". We should indeed highlight what makes Haskell special, but we should do that by highlighting the end result (a better developer experience) rather than the raw ingredients (laziness and purity).

malteneuss:
> Long-term maintainable" is also good....
> I'd prefer "you can rely on" or "rely upon" over "depend on".

"**Enjoy long-term maintainable software you can rely on**"
image

hasufell:
@noughtmare I disagree for the matter of short headlines (not the principle of expressing what developer experience Haskell provides).

We need to be more verbose when we try to express development experience. It simply doesn't fit well into a one-liner without being meaningless.

noughtmare:
I fail to see how laziness and purity are meaningful words to use in the headline, because I don't think many programmers will be that familiar with these terms.

Also, I thought I'd look at some of the competition, and they all seem to favor highlighting developer experience (except arguably typescript):

* Build simple, secure, scalable systems with Go
* Python is a programming language that lets you work quickly and integrate systems more effectively.
* Swift is a general-purpose programming language that’s approachable for newcomers and powerful for experts.  
   It is fast, modern, safe, and a joy to write.
* (Rust) A language empowering everyone to build reliable and efficient software.
* TypeScript is JavaScript with syntax for types.

I think these headlines do a pretty good job at differentiating the languages.

hasufell:
> I don't think laziness and purity are that meaningful to most programmers.

Exactly... and as a programmer, those terms would make me think a moment: "hey, what does that mean?" and I'd scroll down to the features list or go to wikipedia. But when it's just generic buzzwords, I'm not sure I'd be inclined to read further.

malteneuss:
Laziness as one of the technical things that sets Haskell apart is (and should stay) mentioned in the features section further below, so there's no pressure to mention it in the headline. My goal of this PR and https://github.com/haskell-infra/www.haskell.org/pull/338 is to present the (business) benefits/value propositions (or whatever you might call it) first in terms that everyone can understand. As @noughtmare states, laziness (in Haskell's extend) is unfamiliar to most programmers and not something positive (or negative).

tomjaguarpaw (approved):
Thanks!

I approve of this because it's better than the status quo. I'm sure it can be improved yet further, but no need to hold back this improvement on that basis.

aviaviavi (approved):
Thanks @malteneuss! 

This looks good to me. 

I still feel we can continue to improve on this and welcome further suggestions, but in the meantime, let's absolutely ship it!

rebeccaskinner (approved):
This is a really nice user-centric change.

aaronallen8455 (approved):
I agree that there is room for improvement but that this is preferable to the status quo.

angerman:
I'm uneasy about this change. I feel we are losing a description of what haskell is in favour of some unclear
- what does "enjoy long-term maintainable"
- what exactly do we understand with "rely on"
meaning.
I also question the statements validity.

I'll excuse myself from voting on this.

--

enjoy long-term maintainable: haskell provides a strong type system which gives you rigidity around your code, this does come with the benefit of compiler assisted refactoring. I personally do not enjoy refactorings. This rigidity also means you can easily run into snowballing churn when you need to update packages because you need to use a new compiler, because your operating system changed, and you can't use your old compiler anymore, ... I also don't enjoy that.

rely on: I'm not sure what this means, can I not rely on other programming languages? What exactly can I rely on? Are we trying to sell types as formal method assurances here? Can I rely on the compiler to eat my source code if I have a type error?

I'm not much a fan of the current statement either, primarily due to the unclear nature of what "advanced" is supposed to mean, and the lack of "lazy". Ultimately though if the others are happy about this, change I'm not going to stay in the way, and trust their better judgement.

Tritlo:
I have to agree with @angerman here. I like the current short tagline, but maybe we could include a further textbox underneath?

tomjaguarpaw:
I understand this proposal to be an attempt at increasing our marketing appeal, not a deposition we are swearing to in a court of law.  I am judging this proposal on the basis of "will in encourage more people to try Haskell than the status quo?".

hasufell:
> I am judging this proposal on the basis of "will in encourage more people to try Haskell than the status quo?".

And how do we know?

tomjaguarpaw:
In general, marketing is a well-studied field of human endeavour. I'm far from an expert on it, but I'll use my best judgement and take input from those who are better at it than me.

hasufell:
I'm not an expert either, but from.my understanding marketing starts with identifying the target group(s), then understand what the target group(s) responds to, then develop a presentation that evokes that response. 

We haven't even done step 1, so I'm at loss how this is being judged.

malteneuss:
Given the [Haskell Foundation](https://haskell.foundation/)'s mission statement "broadening the adoption of Haskell, by supporting its ecosystem of tools, libraries, education, and research", i suggest to pick (and i picked) the broader programming community at large (including commercial users), and communicate not in terms of features (why should i care that Haskell is using functional programming and is advanced?) but in terms of value propositions.

Regarding measurements i read on the [Haskell.org](https://www.haskell.org/) page that there's a sponsor who could help:

> [Scarf](https://www.scarf.sh/) provides data and insights on the adoption of Haskell in order to support efforts to grow the Haskell ecosystem and facilitate industry support for the language.

I doubt that changing a few paragraphs will have a big impact, but it would be the first step for a mentality shift.

aviaviavi:
> Regarding measurements i read on the [Haskell.org](https://www.haskell.org/) page that there's a sponsor who could help

Reporting for duty 😄🫡. We are already tracking the traffic to the website and have some (slowly) long-running efforts to get better visibility into package downloads, ghc up, etc. What I can say at a high-level:

1. Haskell.org traffic currently decreasing y/y. 
2. There is a big imbalance between the traffic we see to the website and the download stats that hackage reports, implying we are not convincing readers effectively to give Haskell a try. (getting ghcup and hackage info into Scarf will help quantify this more precisely). 
3. There is a significant amount of corporate traffic in the data, so the opportunity in improving (1) and (2) may be large. 


> the broader programming community at large (including commercial users)

Seems reasonable to me. From beginner/students all the way to professional.

aviaviavi:
Good to know! (Whenever we get this all hooked up, we can support much more accurate stats in Hackage.)

Regardless, I'd argue we want to appeal to a broad programming audience.

tomjaguarpaw:
> I'm not an expert either, but from my understanding marketing starts with identifying the target group(s), then understand what the target group(s) responds to, then develop a presentation that evokes that response.
>
> We haven't even done step 1, so I'm at loss how this is being judged.

I'm judging it based on this PR being significantly better on the "talk about benefits not features" scale. That's likely to make it more appealing to almost any target group, and in particular groups that don't encompass the type of people we already have inside the Haskell community.

ninioArtillero:
> I have to agree with @angerman here. I like the current short tagline, but maybe we could include a further textbox underneath?

Perhaps consensus can be reached by this approach. Adding lazy to the current subheadline —so that what Haskell is does not get diluted— and a new text box highlighting developer experience and bussiness benefits/value as proposed.

chreekat:
Just as a point of order, four of seven committee members have approved this PR. https://www.haskell.org/haskell-org-committee/ says that potentially controversial changes "should be brought to the full committee for a vote, potentially with other steps (like a community-wide request for comments)".

Does approval from a majority count as a full committee vote, or does the committee want to pursue other steps?

hasufell:
> This isn't a language definition or a library API so we're not really worried about "breaking an API". I don't see much harm in changing things around a bit until we settle on something we're really happy with.

That makes a lot of sense actually and I hadn't thought about it.

My feeling was that the committee sets a more or less final stance on what headline to use, which is why I found the quick voting procedure a bit frustrating and I also felt that it shuts down further community discussion a bit. In the end it has just been two weeks since the PR was opened, which was barely enough time to consider alternatives.

But I think everyone appreciates the quick turnaround time.

simonmichael:
I just noticed this and related changes on haskell.org - "Enjoy long-term maintainable software you can rely on" and the three subparagraphs. Great work! Haskell's home page is much more appealing than it ever has been.

tomjaguarpaw:
Credit to @malteneuss!

PR #339, opened 2025-05-09
Clarify the site info

tomjaguarpaw:
When @malteneuss was submitting PRs it wasn't clear how to find information about the maintenance of this site: https://discourse.haskell.org/t/emphasize-why-haskell-on-haskell-org-landing-page/12036/35?u=tomjaguarpaw. Hopefully this PR fixes things a bit.

### Before

screenshot

### After

screenshot

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -516,7 +516,7 @@ isHome: true
 
          <div class=" span6 col-sm-6">
             <h3>Psst! Looking for the wiki?</h3>
-            <p>This is the new Haskell home page! The wiki has moved to <a href="https://wiki.haskell.org">wiki.haskell.org.</a></p>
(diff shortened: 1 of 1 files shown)


TikhonJelvis:
Maybe a separate PR, but seems like we could add a link to this repo under "Maintenance"

TikhonJelvis:
Also, now that I'm thinking about it: how long ago did the wiki move? I would not be surprised if it happened before I got involved in the broader Haskell world around 2013 :P

tomjaguarpaw:
Yeah, both things seem like worth tackling as follow-ups.

(I would say the quality of the Wiki is so low we shouldn't be advertising it!)

aviaviavi (approved):
Agree about the potential follow-ups but this looks good to me!

tomjaguarpaw:
Thanks for prompting this, @malteneuss

PR #340, opened 2025-05-09
Remove Equinix as sponsor

tomjaguarpaw:
Move the Scarf entry into its place to avoid leaving a gap

----

We are grateful to Equinix for many years of very generous sponsorship!

@jmct, can you confirm  you think this is the right thing to do?

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -490,7 +490,7 @@ isHome: true
       </div>
       <div class=" row ">
          <div class=" span6 col-sm-6">
-            <p><strong><a href="https://metal.equinix.com/">Equinix Metal</a></strong> provides compute, storage, and networking resources, powering almost all of Haskell.org in several regions around the world.</p>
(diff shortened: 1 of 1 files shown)


gbaz:
We should check with jmct and @davean who else we might want to add as an infrastructure sponsor as well. For example, we should add DigitalOcean, which now provides hosting for hackage.

rebeccaskinner (approved):

tomjaguarpaw:
I intend to merge this PR by Friday unless overruled by @jmct. We can always revert it if necessary, but it seems the right thing to do since Equinix are not actually sponsoring us any more.

Happy also to incorporate new sponsors in this PR or a subsequent one.

davean:
DigitalOcean, Discourse, those are the new ones we should be calling out.
Sad to see Equinix Metal shutdown.

I'm not sure of Galois status ATM.

jmct:
Digital Ocean is on haskell.foundation, but they should also be on here as well.

I've reached out to the Discourse folks about how they want us to proceed.

So, I say merge this as is.

As for Galois... I wouldn't rule out some sort of in-kind infrastructure help in the future (I pretty consistently talk with folks there about the HF and how Galois may help), but it's probably wise to remove for the time being (maybe we replace them with Digital Ocean)

Issue #373, opened 2026-04-06
Proposal for landing page redesign


Zelenya:
I had a markdown sketch lying around for a redesign of the front page since 2023, but the recent [What Would You See Changed in Haskell?](https://blog.haskell.org/what-would-you-see-changed-in-haskell/) rekindled the idea.

**preview:** https://zelenya.com/haskell-org-mvp/redesign/
**tldr:** focus on whys and make explicit paths for different target audiences.

### Main goals and steering points

That blog post brings up a crucial question: *“To whom does Haskell cater?”* Why not be explicit and honest about different use cases and directions? So, I took the suggested Academia, Industry, Hobbyists, and reframed them as verbs. This is one of the main pillars of redesign. 

*It would be valuable to link to detailed guides for different applications/domains, but this requires a bigger effort and can be added to the docs later.*

Currently, the page is focused on **what** Haskell is. I love Haskell, and I'd like to emphasize what Haskell is FOR and WHY others should care (and if they do, guide them to the right places). I've used Haskell in production for years and have read [blog posts about Haskell in production](https://serokell.io/blog/haskell-in-production) for years, so I tried to pull even more towards the main "production" strength:

> Building complex software that I can reason about (and maintain)

This includes things like invariant/domain modelling, fearless refactoring, and reliability. So I tried to organize around those blocks. I might be biased, and this might contradict the *"To whom does Haskell cater?"*, but I think the resulting "whys" are fairly balanced to appeal to all (I mean all who care, still avoiding success-at-all-costs).

### Other decisions

I grouped the initial "why" and "feature" sections under the same section because all those bullet points answer the same questions: "what". However, I organized them by the "path" again (for example, we should emphasize performance for industry but not for hobbyists), and only the packages appear in all (those need to be redone).

I'll be straightforward. "Excellent tooling" is a bit exaggerated and might lose people's trust. It lists minimal requirements for the tooling. Let's be honest, Rust might claim "excellent" tooling, but Haskell can't (see the tooling section in the same blog post or haskell report). I do understand that Haskell offers tools that other languages/ecosystems don't even dream about, but this is not listed here. Haskell is infamous for its tooling, and this requires a much bigger effort to offset. That's the only one I suggest removing.

---

Videos also depend on the chosen path. I took some original videos, suggestions from [this issue](https://github.com/haskell-infra/www.haskell.org/issues/288), and some popular ones from youtube.

### Controversial part

[...]

PR #381, opened 2026-04-25
Home redesign 2: Choose your Haskell Path

Zelenya:
Second phase of https://github.com/haskell-infra/www.haskell.org/issues/373.

> [The blog post](https://blog.haskell.org/what-would-you-see-changed-in-haskell/) brings up a crucial question: “To whom does Haskell cater?” Why not be explicit and honest about different use cases and directions? So, I took the suggested Academia, Industry, Hobbyists, and reframed them as verbs.

- ~~Added four buttons/tabs/pills. **Based on the active selection, the following content changes.**~~
- ~~It also affects the new section with CTA links at the bottom of the page (between features and sponsors).~~
- ~~Note that it defaults to industry (a second button, so it's more clear that those are "clickable")~~

In original proposal, I wanted those CTAs together with the path section, but I got some feedback that it was too crowded and overwhelming

---

After more feedback, simplified, see the comments below.


site/index.html
```diff
@@ -530,6 +586,50 @@ <h3>Packages</h3>
...
+                  <a href="https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/table.html" class="btn btn-default btn-sm">GHC language extensions</a>
+                  <a href="https://github.com/ghc-proposals/ghc-proposals" class="btn btn-default btn-sm">GHC Proposals</a>
+                  <a href="/documentation/" class="btn btn-default btn-sm">Documentation</a>
+               </p>
+            </div>
+            <div class="cta-pane" data-path="hobbyists">
+               <h3 class="path-start-heading text-center">Start building your next fun project.</h3>
+               <p class="path-start-links text-center">
+                  <a href="/get-started/" class="btn btn-primary btn-sm">Get started</a>
+                  <a href="https://www.reddit.com/r/madeinhaskell/" class="btn btn-default btn-sm">r/madeinhaskell</a>
```
tomjaguarpaw: `r/madeinhaskell` seems to have been dead for 5 years and perhaps it was never alive because it only has 20 or so posts, so this is not a good advert for Haskell.

tomjaguarpaw:
For the benefit of those who didn't read the issue ticket #373, @Zelenya has a preview at https://zelenya.com/haskell-org-mvp/redesign/

tomjaguarpaw:
Otherwise this looks good to me

site/index.html
Zelenya: good point, any other suggestions?

Zelenya: Maybe https://serokell.io/blog/best-haskell-open-source-projects?

tomjaguarpaw: Seems good!

Zelenya:
```suggestion
                  <a href="https://serokell.io/blog/best-haskell-open-source-projects" class="btn btn-default btn-sm">29 Awesome Projects</a>
```

tomjaguarpaw (approved):

rebeccaskinner (commented):
I like the general direction that this is going, but I'm not sure I like the way that selecting a path above the fold changes the CTA section toward the bottom of the page, without making it clear that these two things are related. 

In your [MVP](https://zelenya.com/haskell-org-mvp/redesign/#proof) those were co-located. I'm open to pushback, but I think that was better since it would allow someone to explore all the different paths (or at least see that they were related).

Separately, I also wonder if maybe the "Why Haskell" section might not be something we'd eventually want to consider putting behind the paths as well- I could see some value in specializing these to the particular reader, but that's a bigger change and I don't think it would be necessary in this PR.

Zelenya:
@rebeccaskinner I see how this could be confusing. The problem with the mvp is that it was too overwhelming. There are three "layers" of actions, and all of them are different:
Screenshot_2026-04-24_at_8 52 44 AM

On a smaller laptop screen, it's ok because a scroll hides it, but I couldn't make it work for larger screens.

A simple alternative could be to remove the first CTAs below the logo and only have those path-dependent CTAs. I didn't choose that direction because I assume a lot of people come to haskell.org to go to the Get Started page, and it's important to keep it easily accessible.

And yeah, a bigger change could be to move the path selection below the logo, and make it affect "WHY" as well. The path-dependent CTAs can go below whys and give some breathing room (might need to remove the first CTAs anyway). I was thinking the second pillar only because it's the least generic (but we can do more).

What do you think? Should I try removing those and re-grouping things?

abailly:
I also think that we should keep a clear `Get Started` CTA at the top of the page. 
Personally when I try a new language I want 2 things: a good, compact overview of what the language looks like, how it "Feels", and a place to easily get started playing with it, locally or online. The details come after.

Zelenya:
@rebeccaskinner What if we do something like this? 

Screenshot 2026-05-10 at 8 21 21 PM
...
Screenshot 2026-05-10 at 8 21 29 PM


- The middle pillar depends on the select path now
- New "users" can see that buttons affect the content right away
- There is a bit less going on (removed the BIG CTAs from prev. PR)
- `Get Started` is still visible on the first screen (below the main whys)
- Also, for people who want just `Get Started`  
  - There is also `Get Started` first thing in the header,
  - (There will be another `Get Started` at the bottom of the page


_Haven't pushed the code yet, cause I'm not sure I like the background there and need to think about text a bit more_

tomjaguarpaw:
@rebeccaskinner and @abailly, any thoughts?

Thinking through it a bit more I'm hesitant about these buttons changing the content. I'm not a UX expert so I'm happy to be overruled if I'm missing something, but I don't recall seeing a site where content of this sort is hidden and exposed like that. (The [Python front page](https://www.python.org/) has something that functions similarly but it shows code examples rather that more meaningful text.)

Someone could realistically have an interest in Haskell of a sort that means two paragraphs that are relevant to them can only be seen separately. Put another way, I don't see why someone should only be able to see "Build reliable software" if "Build your next pet project" is hidden.

Can we just have all these paragraphs visible at once? Then visitors can scan and focus on whatever catches their eye.

rebeccaskinner:
@tomjaguarpaw I'm of two minds personally.

I agree that the UX is unusual, although I don't think that necessarily needs to be a blocker here. The biggest concern is, to your point, that people might be interested in Haskell for more than one reason, and more broadly I don't think every benefit of the language clearly falls into a particular domain. For example, in the screen shot we see modeling complex domains under "build your next project" but clearly that's of interest to both reliability and language design as well.

The counter-point here is that I think if we put all the paragraphs above the fold we might end up being too text dense and we'd need to move the tutorial repl down lower, or split the content.

My initial proposal is that we could do something like we did with the testimonials, where we collapse them all into a single carousel and order them randomly. We might also want to curate and pick the strongest paragraphs from each section.

Zelenya:
@rebeccaskinner @tomjaguarpaw What do you think of this? 

Screenshot 2026-05-26 at 8 46 23 PM

Note that I moved it to the bottom not to conflict with all the columns at the top. 

---

<details>
  <summary>I don't think that added complexity with carousel worth it (unless we add one or two more things)</summary>
    
  Screenshot 2026-05-26 at 8 28 43 PM
  
</details>

(I've also tried 2x2, but wasn't happy with it)

abailly:
@Zelenya I like your latest proposal, however I think the fact "I like" something does not mean the design will serve its purpose which is to engage people in trying Haskell and use it more for fun and profit (if that's the goal). 
I wonder how complicated it would be to try some kind of A/B testing, or provide users a way to select the design they prefer and decide based on how many people chose one or the other?

abailly:
I have collected a few screenshots from other languages' home page, as a reference of sort: https://photos.google.com/u/1/album/AF1QipP4pMVGSA54njdin_Vy1C8MY-GnBIgkDepMXWiF 
I don't know how sharing works in Google photos so let me know if you cannot see them.

Zelenya:
@abailly It's 404.

abailly:
> @abailly It's 404.

Apologies: this link should work https://photos.app.goo.gl/PakD732o5ZhzMkPW6

Zelenya:
Racket has tabs/selectors: https://racket-lang.org/#mature

tomjaguarpaw:
I'm happy to approve website PRs liberally with the intention of moving fast and cleaning up fast later, if we decide we want to change something. I think that's better than the status quo of taking a long time to decide and thereby losing potential improvements.

So, my approval still stands. @rebeccaskinner, @abailly, what do you think?

abailly:
@tomjaguarpaw Let's move on, _done is better than perfect_

tomjaguarpaw:
OK, let's try to get majority to approve this

* [x] @abailly 
* [x] @rebeccaskinner 
* [ ] @Tritlo 
* [ ] @doyougnu 
* [ ] @aviaviavi 
* [x] @Kleidukos

rebeccaskinner (approved):

tomjaguarpaw:
I'm going to merge this on the basis that it is "non controversial". If that assessment is wrong we can revert it, but I think it's better to act here than delay.

Issue #46, opened 2020-12-11
Outdated cabal documentation


tomjaguarpaw:
The [original Cabal proposal](https://www.haskell.org/cabal/proposal/index.html) (which must be from about 20 years ago) appears on Haskell.org.  It is not clearly marked as an archive of a historical document. [A user tried to follow it to enable `FlexibleContexts`](https://www.reddit.com/r/haskell/comments/hwy8ux/how_to_draw_line_charts_with_interactive_comments/fz337na/), specifically the [The Cabal simple build infrastructure](https://www.haskell.org/cabal/proposal/x611.html) section.

Is it possible to perhaps just delete it and let the internet archive preserve it for posterity? Alternatively, if as a historical document it should be preserved, could it be clearly marked as a historical document, not for up-to-date use, and perhaps even moved to another domain (haskell-history.org?).

tomjaguarpaw:
It's really, really confusing for users to come across documentation that is not up-to-date.  See [an example that occurred just today](https://mail.haskell.org/pipermail/haskell-cafe/2020-December/133139.html).

tomjaguarpaw:
Migrated to https://github.com/haskell/cabal-website/issues/14

PR #383, opened 2026-06-28
Fill (and refill) the video spots

Zelenya:
Second round of https://github.com/haskell-infra/www.haskell.org/issues/288, related to https://github.com/haskell-infra/www.haskell.org/issues/373

Reasons to touch the videos:
1) One video was just removed and we had a free spot.
2) When I look at the current list from the lens of Hobbyist vs Academic vs Industry, most (if not all) are leaning towards the "Academic" audience. 

I wanted to add something for Hobbyist and "more mainstream" audience, I have two candidates: [Haskell in 100 Seconds](https://www.youtube.com/watch?v=Qa8IfEeBJqk) and [Making Music with Haskell From Scratch]. I think either could bring some balance to the rest of the videos (I could also see an argument that this makes haskell.org look "less serious")

But I couldn't choose one and also think that "Monad" video doesn't even have Haskell in the title, so I'd suggest swapping it.

---

I also really like these two:
- [The Haskell Unfolder Episode 25: from Java to Haskell](https://www.youtube.com/watch?v=YwshlQXKO80)
- [Developing an application from scratch (Haskell Unfolder #46)
](https://www.youtube.com/watch?v=5W0ZUY_l1dU)

but there are only 6 spots unless someone makes another carousel or something.

---

Screenshot 2026-06-28 at 4 21 34 PM

### DIFF
--- a/site/index.html
+++ b/site/index.html
@@ -160,6 +160,14 @@ isHome: true
                            </a>
                         </div>
                         <div class=" span3 col-xs-6 col-sm-3 col-md-2">
+                           <a href="https://www.youtube.com/watch?v=Qa8IfEeBJqk" title="Haskell in 100 Seconds" class="thumbnail">
+                              <img src="https://i1.ytimg.com/vi/Qa8IfEeBJqk/mqdefault.jpg" alt="" class="img-responsive">
+                              <div class="caption">
+                                 <h5>Haskell in 100 Seconds</h5>
+                              </div>
+                           </a>
+                        </div>
+                        <div class=" span3 col-xs-6 col-sm-3 col-md-2">
                            <a href="https://www.youtube.com/watch?v=4RuLzL_q0zs" title="Past and Present of Haskell &ndash; Interview with Simon Peyton Jones" class="thumbnail">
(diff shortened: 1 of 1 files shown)


Zelenya:
cc @kosmikus and @tomjaguarpaw (you looked into this before and I would like to know what you think)

tomjaguarpaw:
I think the best thing would be to pick one and go with it. Then once the video list looks more balanced again we can think about making a carousel (or some other approach). My preference would be "from Java to Haskell" but the other is fine by me too.

tomjaguarpaw (approved):

rebeccaskinner (approved):
This is great!

I agree with Tom; I think any video choice works for now and I also like the idea of featuring From Java to Haskell.

Noting this more as a general idea for later, not a change request for this PR: I wonder if it would make sense to handle videos the same way we handled testimonials- a carousel populated from YAML with the order shuffled at load time.

Zelenya:
@tomjaguarpaw could you please clarify? Are you suggesting "From Java to Haskell" instead of the current empty spot, and bringing back the monad video? Or are you ok with removing the monad video and prefer "From Java to Haskell" to one of the suggestions?

tomjaguarpaw:
> @tomjaguarpaw could you please clarify?

I'm OK with all of your suggestions. The merged one is fine.

PR #385, opened 2026-07-04
Made collapsable elements visible when javascript is disabled

### LINKED ISSUE #384: Site isn't navigable when JavaScript is disabled
When JavaScript is disabled (either through the developer console or a browser plugin like noscript) then elements of the website break.

Some elements are reasonably expected to break, such as the playground, however the `.collapse` UI elements also stop working. This renders the hamburger menu impossible to open and the "Installation via native OS package manager" section on the downloads page impossible to open.

It would be nice for the few users who do have JavaScript disabled to still be able to see these elements.

ShaiyaJ:
Closes #384

Originally I opted to put all collapsible sections inside `<details>` tags, but it proved difficult to get the animations right for this.

Instead, I think this is a good middle ground. All elements with the `.collapse` class are set to `display: block;`, and all `[data-toggle="collapse"]` elements are set to `display: none;`.

The result is collapsed areas are visible, and the elements that toggle the collapsing behavior are hidden (this might want to be reverted though). The site is a little less pretty to those with JavaScript disabled, though.

### DIFF
--- a/site/templates/default.html
+++ b/site/templates/default.html
@@ -29,6 +29,17 @@
             } catch (err) { }
           })();
       </script>
+      <noscript>
+        <style>
+          .collapse {
+            display: block;
+          }
+
+          [data-toggle="collapse"] {
+            display: none;
+          }
+        </style>
+      </noscript>
    </head>
    <body class="page-$page$">
      <div class="wrap">


tomjaguarpaw:
Seems like a reasonable approach. Let's give it a go.

rebeccaskinner (approved):

tomjaguarpaw:
Thanks!

PR #387, opened 2026-07-11
Replace Try Haskell with link to Haskell playground

### LINKED ISSUE #386: Try Haskell has shut down
The "Try It!" section of https://www.haskell.org/ no longer works because Try Haskell has shut down. See https://discourse.haskell.org/t/tryhaskell-org-has-shut-down/14389. We should remove that section so the site doesn't appear broken.

tomjaguarpaw:
Try Haskell has been shut down, so that section of the front page is broken before this commit

Fixes https://github.com/haskell-infra/www.haskell.org/issues/386

screenshot

----

We may want to improve this later, but for the moment we need an emergency hotfix so the front page isn't broken.


tomsmeding:
Subjectively having a huge "Try it!" box with just a single line of text with a link in it (which then proceeds to go to a place already linked to twice on the page above that box!) feels a bit like an anti-climax.

There is already an "Open playground" button front and center; it's just deemphasised relative to the button next to it ("Get started"); the "Try it!" block below then draws attention again by being heavily colour-highlighted. Perhaps a different fix is to simply cut out the whole "Try it!" block and make the "Open playground" button also coloured (probably the same colour as "Get started"). You could even relabel it "Try now on the Playground", or something.

This is not intended as a reject review, but since changes like these have a tendency to stick for a longer time than you want (since properly fixing this takes a non-zero amount of work), and it concerns one of the primary eye-catchers on the haskell.org front page, I think it's prudent to consider a bit.

tomjaguarpaw:
Feel free to put up an alternative PR. I'm happy to vote in favour of any PR that removes the brokenness from the front page.

Kleidukos (approved):

tomjaguarpaw:
We did https://github.com/haskell-infra/www.haskell.org/pull/388 instead

PR #388, opened 2026-07-11
Remove the "Try it!" block now that tryhaskell.org is gone

### LINKED ISSUE #386: Try Haskell has shut down

tomsmeding:
This is an alternative to #387.
Fixes #386.

Screenshot 2026-07-11 at 20-39-02 Haskell Language

The page loses some splash by the removal of the big dark block, but the upside is that the page looks more like it was designed to look like this now, instead of looking like "here was a block with neat functionality but it broke so here's a placeholder sentence instead".

I'm okay with #387 but wanted to post this as an alternative solution. I do sympathise with @tomjaguarpaw's note of urgency and to not bikeshed too much on this -- after all, the haskell.org homepage is currently broken.


tomjaguarpaw (approved):

tomjaguarpaw:
I'm happy to go with whichever other members of the committee like (or even one then the other, if a clear preference is established after the fact).

Kleidukos (approved):
This one seems better to me. Thanks a lot @tomsmeding.

